PR #2, opened 2019-06-21
Global naming convention and getters

alexandrebodin:
**Introduce a global naming convetion for services, controllers and models.**

Read the proposal [here](https://github.com/strapi/rfcs/blob/proposal/global-naming-convention/rfcs/0000-template.md)

### DIFF
--- a/rfcs/0000-template.md
+++ b/rfcs/0000-template.md
@@ -1,0 +1,143 @@
+- Start Date: 2019-06-21
+- RFC PR: (leave this empty)
+
+# Summary
+
+- Introduce a global naming convention for controllers/services/models.
+- Introduce getter APIs.
+
+# Example
+
+```js
+'use strict';
+
+module.exports = {
+  find() {
+    // access to a service
+    strapi.service('admin::servicename');
+    strapi.service('plugins::users-permissions.user');
+    strapi.service('app::apiname.servicename');
+    // or shortcut
+    strapi.service('apiname.servicename');
+  },
+};
+```
+
+# Motivation
+
+Currently, accessing a service, a controller or a model when customizing a Strapi application requires knowledge about Strapi's internal structure.
+
+- The main goal of this proposal is to provide users with a clear naming convention on how to access some part of their business logic and plugins' business logic.
+
+- This proposal will remove the naming collisions between apis that exist today.
+
+- This proposal will also allow Strapi's core framework to evolve without breaking users' applications that rely on some internal apis.
+
+# Detailed design
+
+## Naming
+
+Th first step is to define a naming convention so we can use it throughout Strapi's codebase.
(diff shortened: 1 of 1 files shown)


Aurelsicoko:
LGTM 👍 

------

FYI, I tried to rewrite the entire controller User.js of the Users & Permissions plugin, and I found it was difficult to create the string to get a service. It wasn't natural to write :

```js
strapi.service('plugins::users-permissions.user').add(ctx.request.body)
```

but it was easier in my mind to write this instead

```js
strapi.plugin('users-permissions').service('user').add(ctx.request.body);
```

I think the main reason is that every part is dissociated from each other. First, I'm going into my plugin, I select the service I want to use, and then the method to call. Whereas, in the current proposal, the schema of thinking is a bit different.

**Current proposal:**
Call the service > Find the location > Call the method

**My draft proposal:**
Find the location > Call the service > Call the method

alexandrebodin:
> LGTM 👍
>
> FYI, I tried to rewrite the entire controller User.js of the Users & Permissions plugin, and I found it was difficult to create the string to get a service. It wasn't natural to write :
>
> ```js
> strapi.service('plugins::users-permissions.user').add(ctx.request.body);
> ```
>
> but it was easier in my mind to write this instead
>
> ```js
> strapi
>   .plugin('users-permissions')
>   .service('user')
>   .add(ctx.request.body);
> ```
>
> I think the main reason is that every part is dissociated from each other. First, I'm going into my plugin, I select the service I want to use, and then the method to call. Whereas, in the current proposal, the schema of thinking is a bit different.
>
> **Current proposal:**
> Call the service > Find the location > Call the method
>
> **My draft proposal:**
> Find the location > Call the service > Call the method

Interesting thought. In most frameworks there is only one root access to avoid increasing the API surface area. It might be more a matter of habits because you have to do `strapi.plugins[plugin]` right now :).

Anyway as long as we keep global UIDs we can provide plugin scoped **getters** quite easily. Something like that would do:

```js
const ctx = {
  plugin(plugin) {
    return {
      service(key) {
        return ctx.service(`plugins::${plugin}.${key}`);
      },
      controller(key) {
        return ctx.controller(`plugins::${plugin}.${key}`);
      },
      model(key) {
        return ctx.model(`plugins::${plugin}.${key}`);
      },
    };
  },

  // root api
  service(key) {
    // implementation details
  },
  controller(key) {
    // implementation details
  },
  model(key) {
    // implementation details
  },
};
```

derrickmehaffy:
I see @Aurelsicoko's point of view with the old structure for plugins but realistically it's just a matter of mindset. Either way I like the proposal from @alexandrebodin and think it will work the way it was described but documentation will be important here.

Aurelsicoko:
That's what I think too... just another way of thinking. I shared my point of view, it doesn't mean we should implement it, you have a better experience than me on this topic @alexandrebodin

PR #14, opened 2020-05-08
Add RFCS : for quick entry creation in relation fields

edvincandon:
This RFCS adresses https://github.com/strapi/strapi/issues/4948 in a slightly different way.
I came across the same limitation on a strapi project recently and I tackled it myself..

If this feature ever lands in your backlog, I would gladly help :)

Thank you for reading

### DIFF
--- a/rfcs/xxxx-create-option-on-relational-fields.md
+++ b/rfcs/xxxx-create-option-on-relational-fields.md
@@ -1,0 +1,76 @@
+- Start Date: 2020-05-20
+- RFC PR: (leave this empty)
+
+# Summary
+
+When adding a relation from one content-type to another, currently, strapi only lets us select a pre-existing entry in the specified related content-type. This RFCS will try to describe the motivation and possible implementation to allow the creation of content-type entries via relational input fields.
+
+# Example
+The example below shows a basic working example - If no entries match the input field, the user is given the option to **create** the related content-type entry (via the `mainField` attribute).
+
+![enter image description here](https://user-images.githubusercontent.com/11134131/81402093-0ade7e80-9131-11ea-8183-7bf5cc9e7fc8.gif)
+
+It should not involve breaking or changing the current API.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


alexandrebodin:
Adding link to the preview in the description :)

derrickmehaffy:
This looks great, it would be a huge help for stuff like tag management similar to other CMSes.

:+1:

Aurelsicoko:
It looks really great, how do you deal with required fields? It seems to only work for Collection Types without any required and relational fields. Could you confirm this? 

Great job, thank you very much 👏

edvincandon:
> It seems to only work for Collection Types without any required and relational fields. Could you confirm this?

Hey @Aurelsicoko - Yes exactly, in my use-case I didn't have to handle such Collection Types.

I figure this could be handled by scanning the target Collection Type and making sure :
- there are no required fields other than the collection's `mainField`
- the `mainField` is not a relational field (this condition is already met here : https://github.com/strapi/strapi/blob/c90e3eb67bf2c235c2fdd1b5607eda6b649d9f10/packages/strapi-plugin-content-manager/admin/src/components/SettingsViewWrapper/index.js#L106-L126)

If the feature was activated by default (without the user ticking an option in the settings), this could be the condition to allow it ?

derrickmehaffy:
> > It seems to only work for Collection Types without any required and relational fields. Could you confirm this?
> 
> Hey @Aurelsicoko - Yes exactly, in my use-case I didn't have to handle such Collection Types.
> 
> I figure this could be handled by scanning the target Collection Type and making sure :
> 
> * there are no required fields other than the collection's `mainField`
> * the `mainField` is not a relational field (although this condition is already met here : https://github.com/strapi/strapi/blob/c90e3eb67bf2c235c2fdd1b5607eda6b649d9f10/packages/strapi-plugin-content-manager/admin/src/components/SettingsViewWrapper/index.js#L106-L126)
> 
> If the feature was activated by default (without the user ticking an option in the settings), maybe this could be the condition to allow it ?

I could see this being a common usecase with a `manyWay` relation type, but yeah the required fields would be a problem. Maybe we could have some kind of slider or popout/up window if there is more than one field (think how Ghost works? )

PR #15, opened 2020-06-16
Add Schema Load Lifecycle RFC (0005)

diogotcorreia:
Addresses issue https://github.com/strapi/strapi/issues/5972

As requested, here goes my RFC about the schema load lifecycle.

### DIFF
--- a/rfcs/0005-schema-load-lifecycles.md
+++ b/rfcs/0005-schema-load-lifecycles.md
@@ -1,0 +1,120 @@
+- Start Date: 2020-06-16
+- RFC PR: [#15](https://github.com/strapi/rfcs/pull/15)
+
+# Summary
+
+It would be great if we could manually execute code against the schema object for either Mongoose or Bookshelf. That way, better customization would be possible.  
+This would have the added benefit of being able to add compound indexes, for example, which is not possible at the moment while using Mongoose.
+
+Related issue: [#5972](https://github.com/strapi/strapi/issues/5972)
+Related PR: [#6515](https://github.com/strapi/strapi/pull/6515)
+
+# Example
+
+(not according to the implementation below, but just as an example of what could be done with this)
+
+```js
+module.exports = {
+  lifecycles: {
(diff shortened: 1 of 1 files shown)


alexandrebodin:
Hi @diogotcorreia can you add a link to the preview of the RFC in the description plz :)

diogotcorreia:
Done :)

alexandrebodin:
Thank you for this RFC @diogotcorreia. I will review it in details. Just to give you more info, we are planning on refactor the DB layer a lot before the end of year and we will certainly support some sort of schema lifecycles to do any customization. If we were to implement something now be aware we will probably break it for Strapi v4 :)

alexandrebodin:
Thanks for this RFC. 

Here are somme comments:

- I think we should put the schema specific lifecycles outside of the `lifecycles` key like in idea 2
- Here are some lifecycles I think we could have:
  - before the schema is created to change some schema options
  - after the schema is created to use it (generating indexes for ex)
  - some way to update the process of database table creation (future versions)

- I think we should be passing some ctx to those lifecycles: 
  - connection info (client info)
  - orm instance
  - model info

derrickmehaffy:
> some way to update the process of database table creation (future versions)

@alexandrebodin can you expand on this suggestion?

diogotcorreia:
@alexandrebodin 
Updated RFC in response to your comments

alexandrebodin:
Thanks for the improvements ! 

I think the lifecycles arguments should be passed as an object so we can make it evolve without breaking changes and have potentially different infos per connectors.

Other stuff I would like to see (to answer @derrickmehaffy)

There are a lot of possible events. I need to review our process but here is what comes to my mind:

- before / after schema creation (attributes / fileds config, relation)
- before / after model creation (new Schema or new Model in the ORM)
- before / after dbmodel creation
- before / after dbmodel deletion
- before / after dbmodel edition


Once we define the exact events we can define the APIs for each of them

PR #17, opened 2020-08-16
Add Virtual Attribute Type RFC

Tomaszal:
Slightly modified my approach from PR [#7221](https://github.com/strapi/strapi/pull/7221) and created this RFC as requested.

[RFC Preview Link](https://github.com/Tomaszal/rfcs/blob/master/rfcs/xxxx-virtual-attribute-type.md)

### DIFF
--- a/rfcs/xxxx-virtual-attribute-type.md
+++ b/rfcs/xxxx-virtual-attribute-type.md
@@ -1,0 +1,101 @@
+- Start Date: 2020-08-16
+- RFC PR: (leave this empty)
+
+# Summary
+
+It would be very useful to have "read only" attributes that are generated during "read" lifecycles (find, findOne)
+using developer-defined functions. Currently it is possible to do so for the API (by extending the result in the
+lifecycle functions). However, those attributes are not be available in any plugins (most importantly Content Manager
+and Content Type Builder), since they are not defined in the model. By adding a "virtual" attribute type, you could
+add attributes to a model without them being saved to the database, which would make them available to plugins too.
+
+Related issues: [#3131](https://github.com/strapi/strapi/issues/3131)
+
+Related PRs: [#2712](https://github.com/strapi/strapi/pull/2712), [#6433](https://github.com/strapi/strapi/pull/6433), 
+[#7221](https://github.com/strapi/strapi/pull/7221)
+
+Related discussions: [#7173](https://github.com/strapi/strapi/discussions/7173),
+[#7465](https://github.com/strapi/strapi/discussions/7465)
+
+# Example
+
+## model.settings.json
+
+```json
+{
+  ...
+  "attributes": {
+    "first_name": {
+      "type": "string",
+    },
+    "last_name": {
+      "type": "string",
+    },
+    "full_name": {
+      "type": "virtual",
+    }
+  }
+}
+```
+
+## model.js
+
+```js
+module.exports = {
+  virtualAttributes: {
+    full_name(entity) {
+      return `${entity.first_name} ${entity.last_name}`;
+    }
+  }
+}
+```
+
+# Motivation
+
+Currently it is impossible to add custom fields to admin panel that are not written to the database. As shown in this
+simple example, a user might want to show the full name as a field in the admin panel, while having two separate
+fields for first and last names. Right now the only way of doing this is adding an attribute to the model and modifying
+`create` and `update` lifecycles. This works, however it duplicates the data in the database, which is a bad solution.
+
+This would also solve issue [#3131](https://github.com/strapi/strapi/issues/3131) in a much more practical way (in my
+opinion), as this provides much much more freedom than simple formulas.
+
+# Detailed design
+
+The example shown before is pretty straightforward and should be self explanatory, but I will elaborate further. Before
+Strapi returns an entity it should run a function defined in `virtualAttributes` for each attribute of virtual type.
+Something akin to this (pseudo code):
+
+```js
+// attributes - model type attribute definitions
+// entity - model entity object with populated attributes from the database
+// virtualAttributes - object with virtual attribute functions of the model
+
+for (const key in attributes) {
+  if (attributes[key].type === 'virtual') {
+    entity[key] = virtualAttributes[key](entity);
+  }
+}
+```
+
+Things that would need to be implemented (likely missing some):
+
++ Either Database Manager plugin (ideally) or both Database Connector Bookshelf & Mongoose plugins should not save virtual attributes to the database
++ Database Manager plugin should execute `virtualAttributes` functions after `find` and `findOne` queries
++ Virtual attribute type should be implemented in Content Type Builder plugin
+
+# Tradeoffs
+
+Personally I cannot see any real tradeoffs. This seems relatively not too difficult to implement, while adding a very
+much needed feature to Strapi. It does not require any migration or interfere with existing concepts and features of
+Strapi.
+
+# Alternatives
+
+Issue [#3131](https://github.com/strapi/strapi/issues/3131) suggest having formula fields, which I imagine would have
+a similar implementation (or likely more difficult), while being much less flexible.
+
+# Unresolved questions
+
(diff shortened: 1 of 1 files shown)


remi2j:
I like this idea. Virtual types seem like a great way to make data more readable in the content manager.

Here are some other use cases where it would be useful:

* Save a user's birthday, but use a virtual field to get his age
* Save an article's slug, but use a virtual field to get a link to the full URL
* Generate secret preview links for things like [Next.js Preview mode](https://nextjs.org/docs/advanced-features/preview-mode), using the slug or id and a `.env` token
* Save a relation between User and Article models, but use a virtual field to count how many articles a user has. It would also be useful to sort or filter by that field.

I'm also curious about how it would fit with the Content Types Builder. We probably won't ask the user to write a function in the CTB, but it still has to know that this field exists.

Tomaszal:
Yeah those are some nice use cases, should I add them to the RFC?

Regarding how it would fit in the Content type Builder - I think it should be kept pretty simple. Only ask for a field name with no advanced settings. Perhaps also add a link to the documentation explaining what it is / how to use it.

alexandrebodin:
Referencing an article that was initiating prior the the RFC: https://github.com/strapi/strapi/pull/6433

Tomaszal:
@alexandrebodin any updatese on this?

rfcs/xxxx-virtual-attribute-type.md
```diff
@@ -0,0 +1,101 @@
...
+    entity[key] = virtualAttributes[key](entity);
+  }
+}
+```
+
+Things that would need to be implemented (likely missing some):
+
++ Either Database Manager plugin (ideally) or both Database Connector Bookshelf & Mongoose plugins should not save virtual attributes to the database
++ Database Manager plugin should execute `virtualAttributes` functions after `find` and `findOne` queries
++ Virtual attribute type should be implemented in Content Type Builder plugin
```
alexandrebodin, line 85: Adding this to the Content Type builder create a problem: how to add js code to the Model.js file for virtual fields ? 

As the content type builder write files it is easy as long as those files are JSON. We could imagine writing the js code in the Admin panel and keeping it in the json and running an eval but I don't really like this. 

I think the cleaner approach will be to add a template like system to create virtual fields in the content type builder e.g:

```json
"attributes": {
    "first_name": {
      "type": "string",
    },
    "last_name": {
      "type": "string",
    },
    "full_name": {
      "type": "virtual",
      "template": "${first_name} ${last_name}"
    }
  }
```

and If you want to you can configure the field programmatically a bit like you showed.

rfcs/xxxx-virtual-attribute-type.md
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+```
+
+## model.js
+
+```js
+module.exports = {
+  virtualAttributes: {
+    full_name(entity) {
```
alexandrebodin, line 46: I think here we should have getter and setter. This would allow updating database field based on a virtual field.

alexandrebodin (changes requested):
Thanks for this proposal. Very happy if we can see virtual fields coming to strapi. 

That's a really good start :)

In general I would like you give more details about the content manager (can this be displayed in the list view ? how do we handle sorting, filtering ...) Do we display this field in the edit view ?

rfcs/xxxx-virtual-attribute-type.md
Tomaszal: Yeah I don't like the idea of running an eval either. The template string sounds like a good idea though. It could be parsed using regular JS template literals by doing something [like this](https://stackoverflow.com/a/37217166/7353602). I'll add that to the RFC if this sounds good.

rfcs/xxxx-virtual-attribute-type.md
Tomaszal: That's a nice addition, though it doesn't have a lot of use cases (at least not that I can think of). How would it be best to implement in the `model.js` file though? Perhaps have two additional objects (`getters` and `setters`) in the `virtualAttributes` object, or modify the name of the virtual attribute function (`full_name_get` and `full_name_set`)?

Tomaszal:
> In general I would like you give more details about the content manager (can this be displayed in the list view ? how do we handle sorting, filtering ...) Do we display this field in the edit view ?

I don't really see a reason why this couldn't be displayed, sorted or filtered in the list view. I think it should be handled as regular text in there. Perhaps if the perceived 'type' of the virtual attribute for the content manager is a concern there could be an additional field `virtualType`. That way whenever content manager sees a virtual attribute it will handle it as it would handle an attribute of `virtualType`. However, I'm not sure how necessary is this (I haven't looked too deep into the content manager, so you probably know better).

If a setter is added to the virtual attribute functions then yes, this could definitely be displayed in the edit view. However then there is a problem with user being confused if both the original values and the virtual one are displayed. I.e. if user fills in first name and last name, and then fills in the full name, one thing or the other will be overridden.

darron1217:
How about adding it on model’s js file?

For example
on `api/student/models/student.js`
```
module.exports = {
  virtualAttributes: {
    fullName: {
      get(model) {
        return `${model.firstName} ${model.lastName}`
      },
      // It's optional
      set(model, data) {
        const [firstName, lastName] = data.split(' ')
        model.firstName = firstName
        model.lastName = lastName
      },
      // For filter
      query(options) {
        const [firstName, lastName] = options.fullName_contains.split(' ')
        options.firstName_contains = firstName
        options.lastName_contains = lastName
      }
    }
  }
};
```

darron1217:
How about implement it as `custom field type`?
Then we can create `CombinedText` type and define getter, setter, filter handler... etc on custom area.
And then just select field type on Admin Page.

brettwillis:
@Tomaszal here are some of my comments/inputs:
- It would be most flexible to allow `Promise`'s/async functions for generating the value incase there needs to be some async operation involved in calculating the value (dependency on another model value, external API, etc.)
- There is another use-case (detailed below) where the attribute *is* actually written to the database but is never editable (i.e. the value is generated in the **write** lifecycle, not the read lifecycle). In my opinion we should allow for such a case. How would this be implemented in your proposal?

### Use-case: value generated in write lifecycle

Consider a system that sends push notifications to users when an item is published, and it sets a `notifiedAt: Date` attribute with the timestamp of the notification (which is stored in the database). The `notifiedAt` attribute should be visible to the user so they can see if/when a notification was sent, but they can never edit it (i.e. it is "virtual").

This amplifies the need for async operations during calulation (i.e. sending a push notification).

An alternative/simple way of implementing this is just hard-coding such an attribute as readonly in the model's JSON, but it could be integrated into this proposal if it makes sense?

rfcs/xxxx-virtual-attribute-type.md
semiaddict: Templates are nice, but would likely limit what calculations can be done.
Can't the calculations be done via lifecycle hooks ?

SvenWesterlaken:
Any updates on this?

Inaztm:
Any updates on this?

Tomaszal:
Sorry, I have not used Strapi for any projects since writing this RFC, so I do not have the relevant context to update it with the suggestions from the team. I would suggest someone to take over and update the RFC so that the Strapi team can look into it again.

SvenWesterlaken:
No problem @Tomaszal. I was mainly addressing the straps team in order to bring attention to this again as it would ease up a lot of data I would like to add to my models.

alexandrebodin:
> No problem @Tomaszal. I was mainly addressing the straps team in order to bring attention to this again as it would ease up a lot of data I would like to add to my models.

Hi, this feature does look interesting but isn't something on the roadmap for the time being. Feel free to upvote and comment your specific usecase here https://feedback.strapi.io/feature-requests/p/field-type-computed :)

kaaax0815:
Isn't there any way to display multiple fields in the UI for a relation to another entry?

If not, this RFC should be added

PR #19, opened 2020-09-14
Add Categories to the Dynamic Zone components selector

misterdju:
[RFC preview link](https://github.com/misterdju/rfcs/blob/master/rfcs/xxxx-dz-category.md)

### DIFF
--- a/rfcs/xxxx-dz-category.md
+++ b/rfcs/xxxx-dz-category.md
@@ -1,0 +1,41 @@
+- Start Date: (fill in today's date, 2020-09-14)
+- RFC PR: (leave this empty)
+
+# Summary
+
+Modify the component selector in the dynamic zone by adding the categories
+
+# Motivation
+
+We feel like the “category” on the component isn’t being used to it’s potential. When creating complex layout with lots of possible components, you end up with an horizontal scroll that make the experience quite uneasy. 
+
+Below we seperated our blocks in two categories, one for the basic components, and the other for data associations. 
+
+![categories](https://user-images.githubusercontent.com/505236/93095387-54247900-f6a3-11ea-96f5-d32429cc60a1.jpg)
+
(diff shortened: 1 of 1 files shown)


soupette:
I love the idea especially when working with a lot of components from different categories.

derrickmehaffy:
A bit of feedback also, I'm not sure if horizontal scrolling is the best idea, I like the layout that @misterdju showed off of just expanding vertically. (Thank you for the RFC)

maevalienard:
I like your idea! I would suggest a collapse menu for the categories (i.e. default and media in your example). In my opinion, it would be more scalable with a lot of components divided into these different categories, what do you think?

misterdju:
@maevalienard yes i think it is a good idea. 
Here is a mockup using the same UX than repeatable component.
collapse

maevalienard:
Looks super cool to me! Can you confirm both categories and components would be sorted by alphabetical order (in order to follow the Content-types builder behavior)?

Many thanks for your time! 😄

misterdju:
@maevalienard, yes i confirm both categories and components would be sorted by alphabetical order.

maevalienard (commented):
LGTM :)

maevalienard:
Sounds like every reviewer said to go ahead @misterdju, as discussed you can now properly start working on your proposition :)

misterdju:
I have a question regarding the case where only one category is available. 
Do I display only the components or do I still display the category selector?
In the last case, do I open it by default?

maevalienard:
According to the component behavior we're using, we can still display the category selector when there's only one category. If possible, I think it would be a nice idea to have it open by default :)

alexandrebodin:
Hi @misterdju I'm going to merge this RFC so we can pass its status as `Active`

PR #21, opened 2020-11-19
Add plugin namespaces RFC

jorisw:
See **[RFC](https://github.com/strapi/rfcs/blob/ee0387a89a9c72853d2c65c4af8f58ccef6e645d/rfcs/xxxx-plugin-namespaces.md)**.

### DIFF
--- a/rfcs/xxxx-plugin-namespaces.md
+++ b/rfcs/xxxx-plugin-namespaces.md
@@ -1,0 +1,86 @@
+- Start Date: 2020-11-19
+- RFC PR: 
+
+# Summary
+
+Support `@my-org/` type namespace prefixes for plug-ins and middlewares
+
+# Example
+
+Throughout the codebase, the following string matching and/or replacement is done on installed plug-ins:
+
+```js
+const pluginId = pluginPkg.name.replace(/^strapi-plugin-/i, '');
+```
+
+... in order to detect plug-ins and give them a pluginId. By replacing this with:
+
+```js
+const pluginId = pluginPkg.name.replace(/strapi-plugin-/i, '');
+```
+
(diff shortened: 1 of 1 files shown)


alexandrebodin:
Hi @jorisw Can you change the link tu the preview of the rfc insteand :)

jorisw:
Like this?

alexandrebodin:
Thanks for this RFC.

Two points:

- Your RFC don't take into account name collision in the code if we strip the org name from the plugin (e.g strapi.plugins.xxxx) what xxx will it be ? If we have two plugins with the same name in two different orgs ?
- In v4 we will certainly move to a config to enable plugin and not use the npm package directly to offer more flexibility. (to be confirmed) In that case this would be solved pretty easily with some like what (eslint, gatsby ...) does with their plugin config

jorisw:
> Your RFC don't take into account name collision in the code if we strip the org name from the plugin (e.g strapi.plugins.xxxx) what xxx will it be ? If we have two plugins with the same name in two different orgs ?

Actually my proposal deals with that in two ways:
- The short term proposal converts package with name `@my-org/strapi-plugin-lolcat-generator` to plugId `@my-org/lolcat-generator`
- The long term proposal converts package with name `@my-org/strapi-plugin-lolcat-generator` to `{ namespace: '@my-org', pluginId: 'lolcat-generator'}`.

> In v4 we will certainly move to a config to enable plugin and not use the npm package directly to offer more flexibility. (to be confirmed) In that case this would be solved pretty easily with some like what (eslint, gatsby ...) does with their plugin config

Sounds good. Plug-ins indeed should themselves provide their namespace and name on the long run. I've found reading `package.json` to be quite a hacky approach.

When is v4 due?

derrickmehaffy:
> When is v4 due?

I don't believe we have a hard date for that yet.

alexandrebodin:
Hi, going to close this RFC as we addressed this in v4

PR #24, opened 2020-12-16
Custom fields RFC

stafyniaksacha:
See [RFC here](https://github.com/digisquad-io/rfcs/blob/plugins-custom-components/rfcs/xxxx-plugins-custom-components.md)


stafyniaksacha:
Closed due to #40

PR #25, opened 2021-03-30
Folders structure for media RFC

shtelzerartem:
[See rfc here](https://github.com/shtelzerartem/rfcs/blob/master/rfcs/xxxx-upload-plugin-folders.md)
RFC is referred to https://github.com/strapi/strapi/issues/8612
Waiting for you comments on this topic!


alexandrebodin:
Hi, thank you for this RFC. 

At first glance I am wondering why the folder model would be necessary ? 

I think have a simple path key in the File model that would represent a structure path like '/dir1/subDir' would be a simple solution to implement virutal folders in the Admin

It would be farily trivial to list the folders / move files around/ deleting everything in a folder etc...

We could event use this to serve as a prefix on hosted service & in the static server that is serving the files.

derrickmehaffy:
> Hi, thank you for this RFC.
> 
> At first glance I am wondering why the folder model would be necessary ?
> 
> I think have a simple path key in the File model that would represent a structure path like '/dir1/subDir' would be a simple solution to implement virutal folders in the Admin
> 
> It would be farily trivial to list the folders / move files around/ deleting everything in a folder etc...
> 
> We could event use this to serve as a prefix on hosted service & in the static server that is serving the files.

I was wondering the same but it has been requested with specific providers (like AWS) and we already support it on at least one provider.

I think if it doesn't raise the difficulty too much, why not but would probably need to be explored.

alexandrebodin:
> > Hi, thank you for this RFC.
> > At first glance I am wondering why the folder model would be necessary ?
> > I think have a simple path key in the File model that would represent a structure path like '/dir1/subDir' would be a simple solution to implement virutal folders in the Admin
> > It would be farily trivial to list the folders / move files around/ deleting everything in a folder etc...
> > We could event use this to serve as a prefix on hosted service & in the static server that is serving the files.
> 
> I was wondering the same but it has been requested with specific providers (like AWS) and we already support it on at least one provider.
> 
> I think if it doesn't raise the difficulty too much, why not but would probably need to be explored.

alexandrebodin:
Hi, Yes I don't see the need for empty folders if we say they are "virtual". If we allow creating folder only when you do a move action then it is unnecessary. 

For the operations it is actually way simpler than iterating over all the files. We can do very simple DB operations to retrieve the list of folders that should be displayed at anytime. with a simple indexing starting this would be very efficient. On the contrary using a relation for folder is less efficient as it means finding files by a relation which requires more processing

Here are some dumb queries that would do the job.

```sql

-- select files in a folder is super easy
SELECT * FROM test where path LIKE '/sub2%';


-- listing all folder and counts or other aggegations
SELECT path, count(*) FROM test GROUP BY path;


-- super easy to create the file structure with this in the ui
SELECT path, count(*) 
FROM test 
GROUP BY path
ORDER BY path, LENGTH(path);

-- find folder of x depth
SELECT path, count(*), LENGTH(path) - LENGTH(replace(path, '/', '')) - 1 as depth
FROM test 
GROUP BY path
HAVING depth = 0
ORDER BY path, LENGTH(path);
```

PR #28, opened 2021-05-21
Add v1 draft of content type definition file syntax refactor

alexandrebodin:
# Content Type attribute defintion syntax

This PR introduces some changes to the content type definition files.

You can read it [here](https://github.com/strapi/rfcs/blob/v4/model-def/rfcs/xxxx-content-type-definition.md)

### DIFF
--- a/rfcs/xxxx-content-type-definition.md
+++ b/rfcs/xxxx-content-type-definition.md
@@ -81,7 +81,7 @@ const articleDef = {
 };
 ```
 
-This relation is ine one direction onle (from Article to Category). It means you cannot fetch or edit the `articles` form a `category` in the Admin Panel.
+This relation is in one direction only (from Article to Category). It means you cannot fetch or edit the `articles` from a `category` in the Admin Panel.
 
 To make the relation visible in both direction you can use the following setup
 
--- a/rfcs/xxxx-content-type-definition.md
+++ b/rfcs/xxxx-content-type-definition.md
@@ -1,0 +1,112 @@
+- Start Date: 2021-05-21
+- RFC PR: (leave this empty)
+
+# Summary
+
+## Summary
+
+We will introduce an improved version of the content type definition files. In this RFC we are only describing the attributes types and not the options for now. We hope to make them more understandable and usable both manually and programmaticaly (CLI / parsing ...)
+
+## Example
+
+```js
+const articleDef = {
+  attributes: {
+    title: {
+      type: 'string',
+    },
+    seo: {
+      type: 'component',
+      component: 'seo',
+    },
+    body: {
+      type: 'dynamiczone',
+      components: ['txt.image-text', 'txt.rich'],
+    },
+    images: {
+      type: 'media',
+      multiple: true,
+    }
+    category: {
+      type: 'relation',
+      relation: 'manyToOne',
+      target: 'category',
+    },
+  },
+};
+```
+
+## Motivation
+
+The main goal of this change is to simplify readability of the content-type files so developers understand better what they are working with. This will also help make the internal database layer more standard for future evolution.
+
+We will also start reserving some property names for later features (e.g `parsing`, `formating`, `validation`)
+
+## Detailed design
+
+Every attribute **MUST** have a `type` property.
+
+### Medias
+
+To simplify the usage of media attributes we are also simplifying the corresponding content-type attribute.
+
+```js
+const articleDef = {
+  attributes: {
+    images: {
+      type: "media",
+      multiple: true | false,
+    },
+  },
+};
+```
+
+### Relations
+
+The major change is for relations that will be change to use a `type` property of relation and a `relation` property to describe which relation it is.
+
+**Examples**
+
+**Many `Articles` have One `Category`**
+
+```js
+const articleDef = {
+  attributes: {
+    category: {
+      type: "relation",
+      relation: "manyToOne",
+      target: "category",
+    },
+  },
+};
+```
+
+This relation is ine one direction onle (from Article to Category). It means you cannot fetch or edit the `articles` form a `category` in the Admin Panel.
+
+To make the relation visible in both direction you can use the following setup
+
+```js
+const articleDef = {
+  attributes: {
+    category: {
+      type: "relation",
+      relation: "manyToOne",
+      target: "category",
+      inversedBy: "articles",
+    },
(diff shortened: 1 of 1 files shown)


rfcs/xxxx-content-type-definition.md
```diff
@@ -0,0 +1,112 @@
...
+    category: {
+      type: "relation",
+      relation: "manyToOne",
+      target: "category",
+    },
+  },
+};
+```
+
+This relation is ine one direction onle (from Article to Category). It means you cannot fetch or edit the `articles` form a `category` in the Admin Panel.
```
remidej, line 84:
```suggestion
This relation is in one direction only (from Article to Category). It means you cannot fetch or edit the `articles` from a `category` in the Admin Panel.
```

rfcs/xxxx-content-type-definition.md
```diff
@@ -0,0 +1,112 @@
...
+
+To make the relation visible in both direction you can use the following setup
```
remidej, line 86:
```suggestion
To make the relation visible in both directions you can use the following setup
```

derrickmehaffy:

gustavpursche:
Many plugins I've seen for gatsby extend in some way these definitions, so I'm wondering whether it would make sense the prefix internal properties. E.g.:

```js
const articleDef = {
  attributes: {
    title: {
      $type: 'string',
    },
    seo: {
      $type: 'component',
      $component: 'seo',
    },
  },
};
```

This would make it easier for you to introduce new properties in the future, without breaking anything. Otherwise you could provide an interface, which plugins then could use for extending the definitions with their own properties.

alexandrebodin:
> Many plugins I've seen for gatsby extend in some way these definitions, so I'm wondering whether it would make sense the prefix internal properties. E.g.:
> 
> ```js
> const articleDef = {
>   attributes: {
>     title: {
>       $type: 'string',
>     },
>     seo: {
>       $type: 'component',
>       $component: 'seo',
>     },
>   },
> };
> ```
> 
> This would make it easier for you to introduce new properties in the future, without breaking anything. Otherwise you could provide an interface, which plugins then could use for extending the definitions with their own properties. This would also limit the risk of two plugins overwriting the same property ...

We already have a way to avoid plugins from creating conflicts with a specific property for each plugin to have their own scope. It makes it verbose for sure but avoids using special prefix or characters and keep each property with their original meaning.

```js
const def = {
  attributes: {
    title: {
      pluginOptions : {
        pluginName: {
          ...
        }
      }
    }
}
```

gustavpursche:
@alexandrebodin Ah, nice. Didn't know about this. Maybe it would make sense to lint the file for non-core attributes and to output warnings for plugin developers. I think `pluginName` needs some definition though: should it use it's full name (e.g. `strapi-plugin-algolia`) or a shortcut (e.g. `shortcut`). I think this could be a convention, but should be mentioned somewhere in the docs :)

rfcs/xxxx-content-type-definition.md
```diff
@@ -0,0 +1,112 @@
...
+  },
+};
+
+const categoryDef = {
+  attributes: {
+    category: {
+      type: "relation",
+      relation: "oneToMany",
+      target: "article",
+      mappedBy: "category",
```
soupette, line 106: I am not sure to understand the `mappedBy` property, can we have a definition of relation that only uses `mappedBy` on one side?

```js
const articleDef = {
  attributes: {
    category: { type: 'relation', relation: 'oneToMany', target: 'category' }
  }
}

const catDef = {
  attributes: { type: 'relation', relation: 'oneToMany', targert: 'article', mappedBy: 'category' } 
}
```
  }
```

Comment (derrickmehaffy) [reply rfcs/xxxx-content-type-definition.md:106]: I'm also with you at @soupette; @alexandrebodin where did the mappedBy name come from? I'm wondering if something like `targetField` would be a bit more clear?

**Comment by stafyniaksacha** [comment]
Hello,

Even if Strapi v4 will be written in JavaScript, I think it could be a good idea to design this feature with Typescript (or Jsdoc equivalent) :   
- having distinguished interfaces for each attribute datatype _(allow plugins to extend interface definitions)_
- having a function exposed from `strapi` or `strapi-utils` that will resolve defaults values, so we don't have to manage `undefined` in core/plugins


```ts
import { defineModel } from 'strapi'

export default defineModel({
  attributes: {
    title: {
      type: 'string',
    },
    images: {
      type: 'media',
    },
  },
})

/**
 * same as:
 * 
 * export default {
 *   attributes: {
 *     title: {
 *       type: 'string',
 *       unique: false,
 *       required: false,
 *     },
 *     images: {
 *       type: 'media',
 *       multiple: false,
 *     },
 *   }
 * }
 */
```

> note: this is inspired by vitejs `defineConfig` / vue3 `defineComponent` syntax 
> see more here: https://vitejs.dev/config/

this is only sugar syntax, but it enables excellent DX

The idea behind `defineModel` is it takes a configuration object define by user, which allow `undefined` values where it is allowed
(see https://github.com/vitejs/vite/blob/main/packages/vite/src/node/config.ts#L67) and resolve it with a frozen object with all default values set (see https://github.com/vitejs/vite/blob/main/packages/vite/src/node/config.ts#L195)

Using typescript interfaces (and not types) to describes attributes allow plugins to extend their definitions:
```ts
// in strapi-core
interface MediaAttribute {
  type: 'media'
  multiple?: boolean
}

// in custom plugin "strapi-plugin-media-sharp"
interface MediaAttribute {
  format: {
    [ name: string ]: {
      width: number
      height: number
      fit: 'cover' | 'contain' | 'fill'  | 'inside'  | 'outside' 
    }
  }
}
```

we may have to handle a custom hook in the `defineModel` function to allow plugins to extend / check the model

rfcs/xxxx-content-type-definition.md
alexandrebodin: the inversedBy / mappedBy  keys come as a pair. to know which side is owner and make sure tje inverse side is pointing to the valid field name.

This naming comes from several ORMs like doctrine / mikro-orm. They are pretty common in the ORM ecosystem and easy to define. We could find something clearer. We are open to suggestions. 
- `targetField` is incorrect as we don't target a field on the other entity but the entity itself. the field on the other side is just a way to access the relation for the other side.

PR #29, opened 2021-05-24
GRAPHQL API

Convly:
GraphQL API v4
This PR introduces a draft version of the GRAPHQL API in v4.

We want to get the conversation started around this topic to get feedbacks quickly so feel free to ask questions and give ideas :)

You can read it [here](https://github.com/strapi/rfcs/blob/v4/graphql-api/rfcs/xxxx-v4-graphql-api.md)

### DIFF
--- a/rfcs/xxxx-v4-graphql-api.md
+++ b/rfcs/xxxx-v4-graphql-api.md
@@ -3,7 +3,7 @@
 
 # Summary
 
-Here is a first draft of what we would like the REST API to look like in v4.
+Here is a first draft of what we would like the GraphQL API to look like in v4.
 
 # Example
 
--- a/rfcs/xxxx-v4-graphql-api.md
+++ b/rfcs/xxxx-v4-graphql-api.md
@@ -5,36 +5,11 @@
 
 Here is a first draft of what we would like the GraphQL API to look like in v4.
 
-# Example
-
-#### Fetching one entity
-
-**Request**
-
-```graphql
-{
-  document(id: 1, locale: "fr", publicationState: LIVE) {
-    data {
-      id
-      attributes {
-        title
-        locale
-      }
-      meta {
-        availableLocales
-      }
-    }
-  }
-}
-```
-
-# Motivation
+## Motivation
 
 The idea behind those changes is to make the GraphQL API simpler, more powerful and customizable.
 It'll also take advantage of the changes in the database layer (especially the new query engine).
 Finally, it'll make it easier to add features over time.
-
-# Detailed design
 
 ## Overview
 
@@ -56,9 +31,14 @@ Finally, it'll make it easier to add features over time.
       attributes {
         title
         locale
-      }
-      meta {
-        availableLocales
+        categories {
+          data {
+            id
+            attributes {
+                name
+            }
+          }
+        }
       }
     }
     meta {
@@ -85,9 +65,22 @@ Finally, it'll make it easier to add features over time.
       attributes {
         title
         locale
-      }
-      meta {
-        availableLocales
+        categories {
+          data {
+            id
+            attributes {
+              name
+            }
+          }
+          meta {
+            pagination {
+              page
+              pageSize
+              total
+              pageCount
+            }
+          }
+        }
       }
     }
   }
@@ -104,7 +97,7 @@ single value or array of value
 
 ```graphql
 {
-  documents(sort: "sort") {
+  documents(sort: "title") {
     data {
       id
     }
@@ -114,7 +107,7 @@ single value or array of value
 
 ```graphql
 {
-  documents(sort: "-title") {
+  documents(sort: "title:desc") {
     data {
       id
     }
@@ -124,7 +117,7 @@ single value or array of value
 
 ```graphql
 {
-  documents(sort: ["title", "-price"]) {
+  documents(sort: ["title:asc", "price:desc"]) {
     data {
       id
     }
@@ -150,7 +143,7 @@ single value or array of value
 
 ```graphql
 {
-  documents(pagination: { page: 0, pageSize: 10 }) {
+  documents(pagination: { page: 0, pageSize: 10 }) {
     data {
       id
     }
@@ -168,7 +161,7 @@ single value or array of value
 
 ```graphql
 {
-  documents(pagination: { start: 20, limit: 30 }) {
+  documents(pagination: { start: 20, limit: 30 }) {
     data {
       id
     }
@@ -184,13 +177,24 @@ single value or array of value
 
 ### Filtering
 
+**Context**
+
+Two challenges were encountered when writing the RFC for the GraphQL filters syntax.
+
+First, the symbols:
+With the V4, we introduce a new query engine with its own syntax. In this syntax, filters starts with a `$` sign.
+Unfortunately, GraphQL inputs don't allow keys to start with a symbol, thus preventing us to keep the `$` syntax here.
+
+Then, the union types:
+Since we want to allow custom scalar filters (`startsWith`, `contains`, `gt`, ...), we need to be able to pass objects with custom keys within. Ideally, we also wanted to be able to type `field: value` to create a shortcut to `field: { eq: value }`. Unfortunately (again), GraphQL don't allow input union type between scalars & object types. The decision was then to remove the shortcut, keeping only the filters object type for each scalar.
+
 **Parameter** `filters`
 
 **Examples**
 
 ```graphql
 {
-  documents(filters: { name: "test", $or: [{ price: { $gt: 10 }}, { title: { $startsWith: "Book" }}] }) {
+  documents(filters: { name: { eq: "test" }, or: [{ price: { gt: 10 }}, { title: { startsWith: "Book" }}] }) {
     data {
       id
     }
@@ -202,7 +206,6 @@ single value or array of value
 
 - `id`: entity id
 - `attributes`: entity attributes
-- `meta`: entity metadata (available locales)
 
 ### Creating
 
@@ -214,18 +217,9 @@ mutation createArticle {
       attributes {
         title
       }
-      meta {
-        availableLocales
-      }
     }
   }
 }
-```
-
-**Response**
-
-```json
-
 ```
 
 ### Updating
@@ -234,14 +228,11 @@ mutation createArticle {
 
 ```graphql
 mutation updateArticle {
-  updateArticle(params: { id: 1, locale: "fr" }, data: { title: "Hello", relation: 1 }) {
+  updateArticle(id: "1", data: { title: "Hello", relation: 1 }) {
     data {
       id
       attributes {
         title
-      }
-      meta {
-        availableLocales
       }
     }
   }
@@ -252,14 +243,11 @@ mutation updateArticle {
 
 ```graphql
 mutation deleteArticle {
-  deleteArticle(params: { id: 1, locale: "fr" }) {
+  deleteArticle(id: 1) {
     data {
       id
       attributes {
         title
-      }
-      meta {
-        availableLocales
       }
     }
   }
@@ -267,6 +255,10 @@ mutation deleteArticle {
 ```
 
 ## Aggregation and Grouping (Connections)
+
+> /!\ Warning: ETA: after v4.0.0
+>
+> The following specification is likely to change before its implementation after the release of Strapi V4.0.0
 
 > Note: We're thinking about replacing the `values` attribute by a `nodes` attributes in the Connection queries,
 > and we would like to have the opinion of our community members on this topic
@@ -280,7 +272,7 @@ query {
       open {
         key
       }
-    } 
+    }
   }
 }
 ```
@@ -309,7 +301,7 @@ query {
         nb_likes
       }
     }
-    
+
     values {
       categoriesConnection {
         aggregate {
--- a/rfcs/xxxx-v4-graphql-api.md
+++ b/rfcs/xxxx-v4-graphql-api.md
@@ -1,0 +1,328 @@
+- Start Date: 2021-05-24
+- RFC PR: (leave this empty)
+
+# Summary
+
+Here is a first draft of what we would like the REST API to look like in v4.
+
+# Example
+
+#### Fetching one entity
+
+**Request**
+
+```graphql
+{
+  document(id: 1, locale: "fr", publicationState: LIVE) {
+    data {
+      id
+      attributes {
+        title
+        locale
+      }
+      meta {
+        availableLocales
+      }
+    }
+  }
+}
+```
+
+# Motivation
+
+The idea behind those changes is to make the GraphQL API simpler, more powerful and customizable.
+It'll also take advantage of the changes in the database layer (especially the new query engine).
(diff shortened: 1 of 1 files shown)


derrickmehaffy:

rfcs/xxxx-v4-graphql-api.md
```diff
@@ -0,0 +1,328 @@
...
+
+### Filtering
+
+**Parameter** `filters`
+
+**Examples**
+
+```graphql
+{
+  documents(filters: { name: "test", $or: [{ price: { $gt 10 }}, { title: { $startsWith: "Book" }}] }) {
```
gustavpursche, line 193:
```suggestion
  documents(filters: { name: "test", $or: [{ price: { $gt: 10 }}, { title: { $startsWith: "Book" }}] }) {
```

gustavpursche:
Thanks for posting this. I do have a couple remarks / questions:

- `publicationState` do you think it might make sense to make this a string, so in some unknown future, a content type could follow a custom flow, which has more states than "live" and "draft"?

- I'm not sure about the naming of `pageSize`. It took me while to understand what could be the difference to `total` and `pageCount`. I'd propose: `totalPages`, `currentPage`, `itemsPerPage` instead.

- In general I'd be very happy if there could be a way (maybe based on the query) to disable pagination at all, because if you need to query for all items I guess it's more resource intensive to retrieve them page by page. Maybe this could look like `documents(pagination: FALSE) {}`. I do think though, having it enabled by default makes sense. This would also improve the following scenario: 

- Static page generator fetches items of page 1
- Editor saves something in the CMS at the same time
- Static page generator fetches items of page 2, which now has an offset by one

This could also be solved, if the pagination wasn't done by an index, but by a hash, so the content of a page doesn't change.

- Personally I don't like the syntax of `sory` much and do think it would make more sense to stick to `ASC` and `DESC`, eg.: `documents(sort: "title:asc")` or `documents(sort: ["title:asc", "price:desc"])`. I also do like the way Gatsby does it more:

```graphql
{
  documents(
    sort: {
      fields: [title]
      order: ASC
    }
  ) {}
}
```

- In the filtering section you mention this example: `title: { $startsWith: "Book" }`. That reminds me, that it would be super useful, if we could introduce custom filters, to e.g. compare dates, or geo-coded data (thinking about `around`, `closeTo` ...)

Stun3R:
You didn't talked about the schema.json. Will you make some change as it's autogenerated ? (Working on the SDK to generate types based on the schema.json you provide)

rfcs/xxxx-v4-graphql-api.md
```diff
@@ -0,0 +1,328 @@
...
+    data {
+      id
+    }
+  }
+}
+```
+
+### Pagination
+
+**Parameter** `pagination`
```
alexandrebodin, line 137: Add limit -1  or equivalent

alexandrebodin:
> You didn't talked about the schema.json. Will you make some change as it's autogenerated ? (Working on the SDK to generate types based on the schema.json you provide)

We are going to continue export the schema but it will jsut have the new format :) Anything specific you had in mind ?

Stun3R:
> > You didn't talked about the schema.json. Will you make some change as it's autogenerated ? (Working on the SDK to generate types based on the schema.json you provide)
> 
> We are going to continue export the schema but it will jsut have the new format :) Anything specific you had in mind ?

It would be great if you could separate the schema and the operations in 2 specific file: `schema.json | schema.graphql` & `operations.json | operations.graphql` or at least give better name to the interface & type created in the schema 🤔

SalahAdDin:
Do you have any plan to support **Subscriptions**?

DigitalCop:
Do you plan to support behind a firewall/proxy support? Current version does not work due to a CDN request to an external resource. Found a workaround, but still...

alexandrebodin:
Linking an issue discussing the Graphql API and some pain points.

- https://github.com/strapi/strapi/issues/11649

Convly:
> Do you have any plan to support **Subscriptions**?

Little update, a first version of the subscriptions has been implemented in https://github.com/strapi/strapi/pull/11129 and is available in the latest beta

alexandrebodin:
Copy pasting and answering from the issue here :) 

> ### Current pain points(s)
> 1. Defining routes is a pain, especially if compared to what it was like in **V3**

I'm not sure I understand. What do you mean by routes ? queries & mutations ? resolvers ? 

> 2. Managing User Permissions is much harder and convoluted (somewhat related to the first pain point)

Can you explain with a given example ? 

> 3. While extending `Mutation` the way it currently is needed to is not per se Strapi related problem, it would be amazing if Strapi would handle this by simply creating a new resolver (I believe this is somewhat how it worked in **V3**)

I'm lost on this one what do you mean ?

> 4. The response format is not great for those working with say Apollo Client 3 (or other clients with cache normalization), where the _GQL Client_ relies on a unique identifier inside it _GQL type_ to track the given object. So having the `ID` inside the `attributes` would be beneficial. Furthermore the nested `data {}` inside the response it somewhat verbose and not sure if it's great DX, not only for the sake of the proportionally increasing nullity checks with each nested relation, but also since it - _again_ - increases the complexity for using the _GQL Client_ with **cache normalization** (this could be solved with something like [`custom merge functions`](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#the-merge-function), but that surely would not be a good DX. **_Perhaps_** this could be solved by moving the `metadata` inside the actual response, say instead of
> 
> ```graphql
> query {
>   posts {
>     data {
>       id
>       attributes {
>         title
>         ...
>       }
>     }
>     meta {
>       pagination {
>         ...
>       }
>     }
>   }
> }
> ```
> 
> it could be this:
> 
> ```graphql
> query {
>   posts {
>     id
>     title
>     meta {
>       pagination
>       ...
>     }
>   }
> }
> ```

this format would add the pagination meta for each entry here I don't think this is what you wanted to do ?

ScottAgirs:
@alexandrebodin thanks for reposting, very keen on helping in any way I can. Again - sadly, this is a major roadblock for me to switch from V3 to V4, for now, I had to shelf the migration to V4 and keep working with V3.

krstivojevicv:
@alexandrebodin Regarding the point 4:

This seems really nice:
```graphql
query {
  posts {
    id
    title
    meta {
      pagination
      ...
    }
  }
}
```

If you are not comfortable with meta being inside attributes scope you can use this approach (this is how it works in Directus)
```graphql
query {
  posts {
    data {
      id
      title
    }
    meta {
      pagination
    }
  }
}
```

But the first approach seems more convenient and having meta for each entry can actually be an advantage

ScottAgirs:
> @alexandrebodin Regarding the point 4:
> 
> This seems really nice:
> 
> [..]
> 
> If you are not comfortable with meta being inside attributes scope you can use this approach (this is how it works in Directus)
> 
> ```graphql
> query {
>   posts {
>     data {
>       id
>       title
>     }
>     meta {
>       pagination
>     }
>   }
> }
> ```
> 
> [..]

This format would likely still cause headaches, because it involved extra layer of nesting and also the actual collection type does not have an identifier on its root level.

However, the first option where `meta {}` is part of the root `collectionType {}`
The `meta {}` could in turn have its own ID attached to it. That setup would likely make working with it on a client a breeze.

abdonrd:
I'm also more in favor of the flat format.

valdeniomarinho:
Sadly I still prefer the V3. **My opinion** is that V4 is way too much verbose and **not good DX**.

`meta{}, data{}, attributes{}` : **YAGNI** and **DRY**
`flat and simple` : **KISS**

tilman:
> Sadly I still prefer the V3. **My opinion** is that V4 is way too much verbose and **not good DX**.
> 
> `meta{}, data{}, attributes{}` : **YAGNI** and **DRY** `flat and simple` : **KISS**

Totally agree to this! Especially if we access deeply nested collections via relations the code get's very long and unreadable.

abdonrd:
I think that this new structure makes everything much more complex.

With this level of nesting…

- Writing `data` and `attributes` in queries so many times makes them more difficult to read.
- When it comes to consulting the data from the clients, you are forced to go through all these nests, making the logic in JavaScript longer.
- Complicate the cache on systems like Apollo Client: https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids

I would love to see this structure simplified.

Also this is a plugin, I think a new v5 version could even be created with breaking changes while Strapi stays at v4.

santiagok986:
I love V3, simple, easy and quick to use
```
query($id:ID!){
  recipe(id:$id){
    id
    name
    img
    ingredients
    steps
    category{
      id
      name
      slug
    }
    autor{
      username
    }
  }
}
```
V4 :( more complex and useless nested 

```
query($id:ID!){
  recipe(id:$id){
    data{
      id
      attributes{
        name
        img
        ingredients
        steps
        category{
          data{
            id
            attributes{
              name
              slug
            }
          }
        }
         autor{
          data{
            attributes{
              username
            }
          }
        }
      }
    }
  }
}
```
V3 vs V4

Use V3  :) => recipe.name 
Use V3  :) => recipe.category.name

Use V4  :/  => recipe.data.attributes.name
Use V4  :( =>  recipe.data.attributes.category.data.attributes.name

Sepifanovskiy:
New response structure results in unreadable code when working with nested collection. But event shallow responses make code ugly with all unnecessary wrappers.

IsaacDdR:
One of the main purposes of GraphQL is lost with this over-engineered "solution"

Dorifor:
I also think that the new API response is bloated and much harder to read / process in our apps. 
It's now harder to make it works and easier to make mistakes (and lose time / sanity)

It's a bit of a shame honestly, hoping to see it take a better direction in the future

SalahAdDin:
Is there any way to make out the v4 response version and just use the v3?

Thanks

IsaacDdR:
Great! I pay Strapi so we can get 0 answers about this issue. I'm moving away.

Jacob87:
``` typescript
rowData?.attributes?.service?.data?.attributes?.price?.data?.attributes?.author?.data?.attributes?.name
```
😭😭

p4bloch:
Adding a comment to emphasize something I didn't see anyone else raise: the nesting issue becomes unnecessarily verbose when you're handling media files that are declared as single-file types. Here's a real example from our migration attempt:

V3
```gql
query HeaderData {
  siteContent {
    id
    logo {
      url
    }
  }
}
```

V4
```gql
query HeaderData {
  siteContent {
    data {
      id
      attributes {
        logo {
          data {
            attributes {
              url
            }
          }
        }
      }
    }
  }
}
```

I agree this can come up a bit counter intuitive for graphql folks who are familiar with the flexibility Apollo Client provides. This seems inspired on the Relay pagination strategy, which of course works but what I loved about Strapi is that it was **simple by default** and you could always extend it as a regular NodeJS server if you needed to.

Keep up the good work!

stafyniaksacha:
> ```ts
> rowData?.attributes?.service?.data?.attributes?.price?.data?.attributes?.author?.data?.attributes?.name
> ```
> 
> 😭😭

This is a pain to handle in typed languages (like in dart, typescript, rust, go). 
Is there any new on this ?

SalahAdDin:
> Adding a comment to emphasize something I didn't see anyone else raise: the nesting issue becomes unnecessarily verbose when you're handling media files that are declared as single-file types. Here's a real example from our migration attempt:
> 
> V3
> 
> ```graphql
> query HeaderData {
>   siteContent {
>     id
>     logo {
>       url
>     }
>   }
> }
> ```
> 
> V4
> 
> ```graphql
> query HeaderData {
>   siteContent {
>     data {
>       id
>       attributes {
>         logo {
>           data {
>             attributes {
>               url
>             }
>           }
>         }
>       }
>     }
>   }
> }
> ```
> 
> I agree this can come up a bit counter intuitive for graphql folks who are familiar with the flexibility Apollo Client provides. This seems inspired on the Relay pagination strategy, which of course works but what I loved about Strapi is that it was **simple by default** and you could always extend it as a regular NodeJS server if you needed to.
> 
> Keep up the good work!

Oh God, my eyes, my EYES!!!

cumironin:
Rest in V4 is very painfull to use, so much easier in V3

matthill8286:
I have used V4 for a variety of projects and couldn't understand why everything in the GQL land is ....data.attributes, it's clunky, so I have reverted back to use V3 as the flat structure also follows the graphql spec https://spec.graphql.org/October2021/, i'm also not sure extracting the ID! outside attributes will play well with Apollo caching when you have nested queries, but I guess time will tell.

Whats the reason as so far I can find one, unless i've missed as to why the need to change the GQL offering??

IsaacDdR:
I got tired using GraphQL V4 
Saying goodbye to Strapi, this already took forever and Strapi only seems to care about their interests.
Regret using it, now to update my older proyects I have use that Attributes.data bullshit.

IsaacDdR:
You literally ruined Strapi for me with that useless data structure.
Bye 👋

SalahAdDin:
Does anyone try to use [this plugin](https://market.strapi.io/plugins/strapi-plugin-transformer)?

Does anyone on the Strapi Staff aware of these problems?

IsaacDdR:
It's impressive how Strapi won't do anything about this. We're talking to a wall. I trusted Strapi and now I'm f***ed bc of this update.

matthill8286:
Use this plug-in https://market.strapi.io/plugins/strapi-plugin-transformer,
you can use V4 but with the decent V3 graphql response format. None of this
crap they are offering in V4.

santiagok986:
The plugin https://market.strapi.io/plugins/strapi-plugin-transformer, only works in REST, how can I make it work in GraphQL ?

SalahAdDin:
> The plugin https://market.strapi.io/plugins/strapi-plugin-transformer, only works in REST, how can I make it work in GraphQL ?

Really? No way!

derrickmehaffy:
@SalahAdDin 

> Does anyone on the Strapi Staff aware of these problems?

Yes we are aware but functionally we can't change this in v4 as it would be a breaking change and the structure specified is needed for some of our upcoming features (primarily content-versioning)

SalahAdDin:
> @SalahAdDin
> 
> > Does anyone on the Strapi Staff aware of these problems?
> 
> Yes we are aware but functionally we can't change this in v4 as it would be a breaking change and the structure specified is needed for some of our upcoming features (primarily content-versioning)

Is there any other solution for this? Cause this structure is really horrible, we had to keep the **3** version cause the response format.

SalahAdDin:
> @SalahAdDin
> 
> > Does anyone on the Strapi Staff aware of these problems?
> 
> Yes we are aware but functionally we can't change this in v4 as it would be a breaking change and the structure specified is needed for some of our upcoming features (primarily content-versioning)

As @alexandrebodin closed this issue, should we understand it will be a solution soon? or won't it be fixed never?

alexandrebodin:
Hello @SalahAdDin,

We closed this RFC as this was already implemented and is not the place for issue management. 

Please do comment in the forum thread if you want to add any suggestions there. We did share the reasons we need this format in the mid and long term there, this might give you the information you are looking for. If you find some solutions that we could use while addressing our constraints we would be glad to hear about them!

PR #31, opened 2021-07-01
[V4] Query engine

alexandrebodin:
# Query Engine

This PR introduces a draft version of the Query Engine.

You can read it [here](https://github.com/strapi/rfcs/blob/v4/query-engine/rfcs/xxx-v4-query-engine.md)

### DIFF
--- a/rfcs/xxx-v4-query-engine.md
+++ b/rfcs/xxx-v4-query-engine.md
@@ -100,7 +100,7 @@ const entries = await db.query("article").findMany({
 });
 ```
 
-### `findWithCount(params)` => `[Entry[], number]`
+### `findManyWithCount(params)` => `[Entry[], number]`
 
 Finds and counts entries matching the params
 
@@ -118,7 +118,7 @@ Finds and counts entries matching the params
 **Example**
 
 ```js
-const [entries, count] = await db.query("article").findWithCount({
+const [entries, count] = await db.query("article").findManyWithCount({
   select: ["title", "description"],
   where: { title: "Hello World" },
   orderBy: { title: "DESC" },
@@ -655,12 +655,33 @@ const entries = db.query("article").findMany({
 
 With this new database layer, relations and components have a unified API for populating them.
 
-You can populate by passing an array of attribute names
+You can populate by passing an array of attribute names.
 
 ```js
-db.query("article").findMany({
-  populate: ["componentA", "relationA"],
+const entry = await db.query("article").findMany({
+  populate: ["category", "tags"],
 });
+
+/*
+{
+  id: 1,
+  title: 'Hello world',
+  category: {
+    id: 1,
+    name: 'Category',
+  },
+  tags: [
+    {
+      id: 1,
+      name: 'Tag 1' 
+    },
+    {
+      id: 2,
+      name: 'Tag 2' 
+    }
+  ]
+}
+*/
 ```
 
 For more advanced usage you can pass an object
@@ -668,9 +689,8 @@ For more advanced usage you can pass an object
 ```js
 db.query("article").findMany({
   populate: {
-    componentB: true,
-    dynamiczoneA: true,
-    relation: someLogic || true,
+    category: true,
+    tags: true,
   },
 });
 ```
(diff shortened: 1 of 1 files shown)


rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
+- Start Date: 2021-07-01
+- RFC PR: (leave this empty)
+
+# Summary
+
+In v4 we are build a new database layer to support the futur evolution of Strapi and we are introducing a new query engine with it.
```
gustavpursche, line 6:
```suggestion
In v4 we are build a new database layer to support the future evolution of Strapi and we are introducing a new query engine along with it.
```

rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
...
+
+The query engine should mostly be used by plugin developers, developers adding custom business logic to their applications and the Strapi core team.
+
+# Example
+
+```js
+qb.query("article").findMany({
```
gustavpursche, line 13:
```suggestion
db.query("article").findMany({
```

rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
...
+  },
+  populate: {
+    category: true,
+  },
+});
+```
+
+# Motivation
+
+We want to make sure we have a more flexible and powerfull database layer to build the futur features of Strapi.
```
gustavpursche, line 28:
```suggestion
We want to make sure we have a more flexible and powerful database layer to build future features of Strapi onto.
```

rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
...
+
+Moreover we want to provide a better experience and more possibilities for plugin developers and application developers to build advanced business logic into their plugins or applications.
+
+This layer is really made for database queries and is at a lower level than our previous version. This layer isn't abstracting a lot of things so we can have full control.
```
gustavpursche, line 32:
```suggestion
This layer is really made for database queries and is at more low-level than our previous version. This layer isn't abstracting a lot of things so developers can have full control.
```

rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
...
+
+# Detailed design
+
+## Query API
+
+To work with entities in the database you can user the Query api:
```
gustavpursche, line 38:
```suggestion
To work with entities in the database you can use the Query api:
```

rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
...
+```js
+const entries = await db.query("article").findMany({
+  select: ["title", "description"],
+  where: { title: "Hello World" },
+  orderBy: { title: "DESC" },
+  populate: { category: true },
+});
+```
+
+### `findWithCount(params)` => `[Entry[], number]`
```
gustavpursche, line 103: Should this method be called `findManyWithCount` to be more consistent?

rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
...
+### `delete(params)` => `Entry`
+
+Deletes one entry and returns it
+
+**Params**
+
+| Param    | Type                           | Description                                |
+| -------- | ------------------------------ | ------------------------------------------ |
+| select   | `string[]`                     | Select some attributes to return           |
+| populate | [`PopulateParam`](#populating) | Select which relations should be populated |
```
gustavpursche, line 184: Does `populate` make sense for delete queries? I'm not sure what I'd expect from that ... does it control the `cascade` behavoir?

rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
...
+// 12
+```
+
+### Avg,Min,Max,Sum
+
+> The rest of this API (avg,min,max,sum,groupby) will specified at a later time.
+
+## Query Builder
+
+When in need fore more control over the queries than with the Query API you can use the underlying query builder.
```
gustavpursche, line 307:
```suggestion
When you are in need for more control over the queries the Query API generates, you can use the underlying query builder.
```

rfcs/xxx-v4-query-engine.md
alexandrebodin: This has the same behaviour create & update to be consistent

rfcs/xxx-v4-query-engine.md
```diff
@@ -0,0 +1,744 @@
...
+    dynamiczoneA: true,
+    relation: someLogic || true,
+  },
+});
+```
+
+You can also apply where filters and select or populate nested relations
+
+```js
+db.query("article").findMany({
```
alexandrebodin, line 681: NOTE: Extends the example a bit more

brettwillis:
I like that population can be controlled for components etc, where currently only population of the top level model can be controlled.

Here are some initial comments:
1. I can see some filtering operators that not all database layers will support (e.g. `$endsWith`, `$contains`, and also complex combinations of and/or). Have you considered if there shall be some defined behaviour when a database connector doesn't support a query? Just throw an error?
2. Some database layers support pagination/continuation via continuation tokens. Tokens could be any type but typically a string. Have you considered supporting continuation tokens from previous queries, as an alternative to `offset`?


Some motivation behind point 2: I'm thinking of some database layers where use of `limit` can be costly, costing a "read operation" for every skipped entity. Pagination in the admin dashboard could be more effective with continuation tokens.

PR #45, opened 2022-08-06
Fix some typos in the template and move the `Example` section

WalkingPizza:
Fixed some typos in the `xxxx-template.md` file.

I also moved the `Example` section to the area between `Detailed design` and `Tradeoffs`, because I believe it is much easier to understand the examples once you have a clearer overview of the proposed design and of the motivation behind the PR.

### DIFF
--- a/xxxx-template.md
+++ b/xxxx-template.md
@@ -12,27 +12,27 @@ If the proposal requires changes to the current API or the creation of new ones,
 # Motivation
 
 Please make sure to explain the motivation for this proposal. 
-It means explaining the use case(s) and the fonctional feature(s) this proposal is trying to solve. 
+It means explaining the use case(s) and the functional feature(s) this proposal is trying to solve. 
 
-Try to only talk about the intent not the proposed solution here.
+Try to only talk about the intent, not the proposed solution here.
 
 # Detailed design
 
-Describe the proposal in details:
+Describe the proposal in detail:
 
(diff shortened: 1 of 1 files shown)


gu-stav (approved):
Thank you!

alexandrebodin:
Hey, 
the idea (based on React rfc) si to actually get a quick glance at what it will look like before the detailed designs where you should have more comprehensive examples :)

gu-stav:
@alexandrebodin Oh, sorry I was too quick then. @WalkingPizza could you move the examples section up again then? Otherwise I can take care of it ...

WalkingPizza:
> Hey, 
> the idea (based on React rfc) si to actually get a quick glance at what it will look like before the detailed designs where you should have more comprehensive examples :) 

Didn't know that, feel free to keep the section where it was if you think it makes more sense!

I only decided to include that change because while writing an RFC, I realized that:
1) I am never 100% sure about the proposed design and keep revising it as the first step of my process.
2) Most examples need captioning and/or explanation if not preceded by an outline of what you are trying to achieve, because oftentimes they are only snippets from a much bigger change.

Anyway, I don't mind either way, in the end I can always decide to write one section first and go back up later on and fill the other.

alexandrebodin:
No biggy I don't have strong opinions we can try it that way and see if we want to revert and as you said someone can adapt the template a little to make the RFC more readable anyway :)

PR #46, opened 2022-08-08
Select fields by omission

WalkingPizza:
You can find the proposal [here](https://github.com/WalkingPizza/rfcs/blob/fields-exclude-parameter/rfcs/xxxx-fields-exclude-parameter.md).

### DIFF
--- a/rfcs/xxxx-fields-exclude-parameter.md
+++ b/rfcs/xxxx-fields-exclude-parameter.md
@@ -1,0 +1,46 @@
+- Start Date: (fill in today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+
+# Summary
+
+When using `fields` in queries, users should also be able to use an `$omit` parameter so they don't have to explicitly list all of the parameters they want, but only those they do not want.
+
+# Motivation
+
+In situations like [this one](https://github.com/strapi/strapi/issues/13945), users might want to keep most of the information from a populate relation, but not all of it.
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


derrickmehaffy:
Really love this idea but it would diverge our abstraction between the entityService and db query layer.

If it's possible to do with native Knex `select` (which is what we are abstracting) then we should go that way but I like the idea to use lodash in the abstraction and still just convert it to a select.

:+1:

alexandrebodin:
Thanks for opening this RFC @WalkingPizza.

The main concern with omits in APIs is that the returned data is not predictable and can create problems with caching/respoonse size and such.

When you omit something but the API evolves you will start receiving new data that you might not want which could add a lot of overhead to your queries or break caching. This is why the preferred approach is only to select what you need and not the other way around. 

I don't think this will make it to the core features but something I would like is to offer some extensibility to our query "language" for user specific usecases that wouldn't be in the core

WalkingPizza:
> Thanks for opening this RFC @WalkingPizza.
> 
> The main concern with omits in APIs is that the returned data is not predictable and can create problems with caching/respoonse size and such.
> 
> When you omit something but the API evolves you will start receiving new data that you might not want which could add a lot of overhead to your queries or break caching. This is why the preferred approach is only to select what you need and not the other way around.
> 
> I don't think this will make it to the core features but something I would like is to offer some extensibility to our query "language" for user specific usecases that wouldn't be in the core

I am sure this feature would have helped query-writers including myself quite a bit, but the points you bring up make total sense, so I guess we'll survive without this.

Looking forward to a more extensible query language in the future!

PR #48, opened 2022-12-01
add component at any location in dynamic zone

AdilAmanat:
Link to markdown: https://github.com/AdilAmanat/rfcs/blob/main/rfcs/xxxx-template.md

### DIFF
--- a/rfcs/xxxx-template.md
+++ b/rfcs/xxxx-template.md
@@ -1,0 +1,34 @@
+- Start Date: 2022-12-01
+- RFC PR: (leave this empty)
+
+# Summary
+
+Make it easy to add new component anywhere in between dynamic zone
+
+# Example
+
+If the proposal requires changes to the current API or the creation of new ones, add a basic code example.
+
+# Motivation
+
+Making it easy for content manager to add new component in between of the dynamic zone. Currently it's very difficult to add a new component in the middle or start of the dynamic zone if the dynamic zone contains large list of components. For example if the dynamic zone has 200 components and you wanted to add new component at 10th index then you have to add component at 201th index and then move it up one by one to place it on the required location.
+
+![Strapi New Component in Dynamic Zone](https://user-images.githubusercontent.com/21098295/205133227-e8fd95ed-736d-439c-81f1-086e6a389cf4.png)
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


AdilAmanat:
@alexandrebodin  can you have a look at it?

alexandrebodin:
Hi @AdilAmanat  I do like the proposal of being able to add a components below or above another directly. I'll ping the Product team on our side to comment on the RFC.

gu-stav:
Just a heads-up here: we recently started to implement a unified way to handle drag and drop (for relations), which we now start to apply to repeatable components and soon dynamic zones. It's not the same as inserting a component at a certain position, but it will allow you to move them around more easily.

You can follow along the progress e.g. here: https://github.com/strapi/strapi/pull/15079

AdilAmanat:
@gu-stav  for the initial thought DND seems to be a fine idea for most of the cases where the dynamic zone doesn't have hundreds of components. But when dynamic zone has 200-300 components, DND would still be a pain point because it wouldn't be easy to add component on the bottom and scrolling it to the top/middle of the dynamic zone to drop it to a required state. 

**Why I think DND is not an eventual problem solver**: For a large dynamic zone where we have 200-300 components in the dynamic zone on Drag, it would be rerendering the entire list on drag and rerendering a large list of UI can result in a laggy experience. To prevent that you would have to apply virtualization on the dynamic zone(which is the optimum idea) but it would have a different UX.

joshuaellis:
Hi everyone, we've picked this up and are moving it into our review process to resolve this RFC. When there's an update on the product decision, we'll update you all.

AdilAmanat:
@joshuaellis do let me know if I can contribute because I have already done the POC and the draft PR is ready.

joshuaellis:
> @joshuaellis do let me know if I can contribute because I have already done the POC and the draft PR is ready. 
> 
> 
> 
> 

I think we'll probably go for an in-house design solution that aligns with the overall product vision. Thank you for sharing a POC and this RFC however, it's always helpful to know how we can improve the content experience 😊

joshuaellis:
Thank you for your patience all, there's a PR open in the main strapi repo with an implementation to do this.

AdilAmanat:
@joshuaellis  thanks for the update.

PR #52, opened 2023-09-04
[V5] Database

Marc-Roig:
This PR introduces a draft version of the database layer changes for the V5 major release.

You can read it [here](https://github.com/strapi/rfcs/blob/v5/database/rfcs/xxxx-v5-database.md).

### DIFF
--- a/rfcs/xxxx-v5-database.md
+++ b/rfcs/xxxx-v5-database.md
@@ -1,0 +1,155 @@
+- Start Date: (fill in today's date, 2023-09-04)
+- RFC PR: (leave this empty)
+
+# Summary
+
+We need to support development of new features. Most of them are going towards a more CMS like model. to support this approach we will introduce the notion of `documents` & `links`.
+
+# Motivation
+
+Bringing new features that were not possible in v4. The most important ones:
+
+## Draft and Publish
+
+Introducing a new version of draft & publish. One which allows having one draft and one published entry at the same time.
+
+This version should support having different relations between a draft and a published version.
+
+Consider a scenario where a writer wants to work on an article (draft) while the previously published version remains unaffected. This requires a system that supports simultaneous draft and published entries, leading to the idea of `Documents`.
+
+## Synchronized relations across locales
+
+In Strapi V4, users requested the ability to share the same relation across multiple locales. This led to the concept of `Links`, allowing relations to be shared across different document locales. And also unblocks other use cases mentioned in this RFC.
+
+Instead of breaking v4 relations `Links` will be a different type, and v4 relations will be kept as legacy relations.
+
+# Detailed design
+
+Documents are introduced in v5 as a way to group multiple entries. That is implemented by adding a new `document_id` column in entries:
+
+![Untitled]()
+
+Also, all the variations of that document will have its own column in every entry:
+
+- `locale` for i18n
+- `publicationState` for draft and publish
+
+## Document id and entry id
+
+Additionally, in V5, all identifiers (entry ids and document ids) will now be universally unique identifiers (UUIDs). This transition offers significant benefits:
+
+- `Data Transfer`: Previously, transferring data between databases while retaining the same entry IDs was not possible, due to the use of incremental ids.
+
+The switch to UUIDs, which will keep entry and document ids unique, makes this feasible.
+
+- `More secure`: The use of UUIDs also enhances security. With incremental IDs, one could predict the existence of other IDs based on a known ID. For instance, if ID `10` is known, it is reasonable to infer that IDs `9`, `8`, etc., exist in the database. UUIDs, being non-sequential and randomly generated, prevent such easy discovery.
+
+While the probability of two UUIDs colliding is not zero, it is practically negligible, making them an excellent choice for unique identifiers.
+
+## Links
+
+`Links` introduce the following key functionalities:
+
+### Relate entry to document
+
+In V4, you might have encountered this situation:
+
+![Untitled]()
+
+When relating an article to a specific author in V4, only articles and authors of the same locale could be linked, because localizations were not really grouped in a single document.
+
+Thus, if an author did not have a locale that matched the article's, creating a relationship was impossible.
+
+`Links` will solve this by referencing a document as a whole, not a specific locale:
+
+![Untitled]()
+
+---
+
+### Synchronize links across locales
+
+In V4, if you want to keep the same relation across the same locales, one would have to manually create them, with the risk of missing one.
+
+![Untitled]()
+
+In V5, `links` will offer the possibility to be synchronized across locales.
+
+![Untitled]()
+
+As an additional comment, the synched relation will only affect the locales inside the same publication state. Meaning, relations will not be shared between the `draft` and `published` versions of the document.
+
+![Untitled]()
+
+---
+
+### Polymorphic links
+
+V5 will eventually support polymorphic links.
+
+Those will allow use cases like Website Menus , linked to many other content types
+
+![Untitled]()
+
+This was just not possible in V4.
+
+## Link content schema
+
+Links will be represented as so in the content type schema:
+
+```jsx
+{
+  attributes: {
+    “myLink”: {
+       “type”: "link",
+			 // Target one (or multiple) content types.
+			 // Target all by not defining target
+       “target”: "uid-a" | ["uid-a", "uid-b"] | undefined,
+       “multiple”: true|false
+		}
+	}
+}
+```
+
+## Links Database Schema
+
+The database schema of `links` join tables :
+
+![Untitled]()
+
+The join table will now contain:
+
+- `Owner side doc id`: Reference to do the document id of the owner side (Article in the example)
+- `Inverse side doc id`: Reference ot the document id of the inverse side (Author in the example)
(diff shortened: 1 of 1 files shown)


rfcs/xxxx-v5-database.md
```diff
@@ -0,0 +1,169 @@
...
+
+We need to support development of new features. Most of them are going towards a more CMS like model. to support this approach we will introduce the notion of `documents` & `links`.
+
+# Motivation
+
+Bringing new features that were not possible in v4. The most important ones:
+
+## Draft and Publish
+
+Introducing a new version of draft & publish. One which allows having one draft and one published entry at the same time.
```
svrin, line 13: I think it should be possible to have more than 2 publication states. Think about having diffrent environments where you want to publish a collection type.

rfcs/xxxx-v5-database.md
```diff
@@ -0,0 +1,169 @@
...
+Documents are introduced in v5 as a way to group multiple entries. That is implemented by adding a new `document_id` column in entries:
+
+<p align="center">
+	<img src="https://github.com/strapi/rfcs/assets/20578351/4a02817e-4004-45ee-9e4a-a755e66c2064" width="300px" />
+</p>
+
+Also, all the variations of that document will have its own column in every entry:
+
+- `locale` for i18n
+- `publicationState` for draft and publish
```
svrin: please be consistent with camelCasing, like publicationState, but document_id.

rfcs/xxxx-v5-database.md
```diff
@@ -0,0 +1,169 @@
...
+  }
+}
+```
+
+## Links Database Schema
+
+The database schema of `links` join tables :
+
+<p align="center">
+	<img src="https://github.com/strapi/rfcs/assets/20578351/f6ebb23e-c6e7-47b6-bdae-8030c80786a6" width="700px" />
```
svrin: pivot_publicationState should be pivot_publication_state. And i think the names should not be prefixed.

rfcs/xxxx-v5-database.md
```diff
@@ -0,0 +1,169 @@
...
+
+Benefits of this approach:
+
+- Links schema reduces the number of queries needed to read and create relations.
+- It is conceptually easy, and similar to our approach.
+- It's easy to share the same relation between different document locales.
+
+Downsides:
+
+- Links will not allow for bidirectionality. A relation that references an entry to a document introduces many complexities that prevents bidirectionality, or makes it rather difficult. The first versions of links will not have bidirectionality in mind.
```
svrin, line 159: I think bidirectional should be included in v5, Otherwise parent/children relationsship of pages collection types are hard to implement in v5.

rfcs/xxxx-v5-database.md
```diff
@@ -0,0 +1,169 @@
...
+
+# Unresolved questions
+
+- How do we manage life cycles?
+  - A database lifecycle would trigger at an entry level, not a document level. As everything will be document oriented, database life cycles might have less sense from a user perspective in V5.
```
svrin, line 164: There should be a method to have lifecycles on document level as well.

derrickmehaffy:
I am quite worried that we won't be able to cover the use-cases properly for lifecycles. Within v4 lifecycles are the only means to directly impact both the content and admin APIs at the same time and we already have limitations around the relational data that doesn't exist within the lifecycles (not even to mention components and dynamic zones which don't have any lifecycle capability in v4 period).

Lifecycles IMO are quite important and something we should take a decision on before stable release and could prevent adoption and migration to v5 for those with more complex use-cases.

Plugin developers also do rely on our dynamic lifecycle injection in v4 quite often and many of our customers are actively using them as well for various reason, down to simple validation that we don't cover normally, all the way up to things like live updating data from external sources and data mirroring.

derrickmehaffy:
For example, lifecycle hooks for components alone has 75 votes: https://feedback.strapi.io/feature-requests/p/lifecycle-hooks-for-components

pwizla:
Minor point here, but on the following schema, shouldn't `article_document_id` from the `article_author` table link to `document_id` and not `id` in the `article` table, or did I misunderstand something? 🤔

Screenshot 2023-09-22 at 11 20 37

Marc-Roig:
@pwizla good catch! Updated the diagram

Marc-Roig:
>I am quite worried that we won't be able to cover the use-cases properly for lifecycles. Within v4 lifecycles are the only means to directly impact both the content and admin APIs at the same time and we already have limitations around the relational data that doesn't exist within the lifecycles (not even to mention components and dynamic zones which don't have any lifecycle capability in v4 period).
>
> Lifecycles IMO are quite important and something we should take a decision on before stable release and could prevent adoption and migration to v5 for those with more complex use-cases.
>
> Plugin developers also do rely on our dynamic lifecycle injection in v4 quite often and many of our customers are actively using them as well for various reason, down to simple validation that we don't cover normally, all the way up to things like live updating data from external sources and data mirroring.

@derrickmehaffy The idea is not to remove the database lifecycles, but to add new ones on the Service layer. Ones where you can hook to listen to document changes.
The database lifecycles will still be maintained and what we could do is an evaluation of which issues there could be on db lifecycles when introducing documents.

At the moment I can not think of any, and we plan to internally use the db lifecycles too! 
Do you think we might want to elaborate on this more? @derrickmehaffy

rfcs/xxxx-v5-database.md
Marc-Roig: Indeed! We will have a document service (similar to the current entity service) with it's own lifecycles.

rfcs/xxxx-v5-database.md
Marc-Roig: We are trying to work on this, and how to make them work with documents. We will keep you all updated. 🙏

rfcs/xxxx-v5-database.md
Marc-Roig: Could you elaborate more on the example? What would be the new publication states in those cases?

One additional feature we had in mind is private drafts, but that's something that might come after having the v5 foundations ready.

Boegie19:
@Marc-Roig Would the end of the entity-service layer not be better? since you could change the service its entity-service request or have entity-service decorators who change stuff. so the only place where you could do this the right way is the end of the entity-service or right of the end of the document service if the entity-service gets removed.

Marc-Roig:
What benefit would you see of having this in the ES rather than on the database layer?
Leaving lifecycles in DB is a must in V5 imo, as we aim to keep as much compatibility as possible.

Boegie19:
@Marc-Roig 
> What benefit would you see of having this in the ES rather than on the database layer? 

ES layer also knows about components and dynamic zones. so has more context. and this would mean if you blocked a request it would also block the components. and not only the highest level of the content-type.

> Leaving lifecycles in DB is a must in V5 imo, as we aim to keep as much compatibility as possible.

  Understandable I am not saying to remove them I am saying maby add a second layer of them.

derrickmehaffy:
> Do you think we might want to elaborate on this more? @derrickmehaffy

Absolutely we need to, both in the changes we will make and the differences to v4 comparatively. I think some of the issues around lifecycles is we don't use them enough ourselves so leveraging them for our purposes would be a good test point.

derrickmehaffy:
> What benefit would you see of having this in the ES rather than on the database layer? Leaving lifecycles in DB is a must in V5 imo, as we aim to keep as much compatibility as possible.

Agreed but I also agree with boegie here around implementing ES lifecycles too so each user can pick the right spot for their code to achieve what they want.

rfcs/xxxx-v5-database.md
svrin: Hi the setup in the case is like having multiple tenant and a collection that gets pubslihed to their different environments:

For example we have consultant that can have multiple web pages where they would be available. 

consultant
| id | name | publication_state | publication_id |
| 1 | Karl Heinz | draft | 0 |
| 2 | Karl Heinz | live| 1 |
| 3 | Karl Heinz | live:immobilien | 1 |
| 4 | Karl Heinz | live:beratung | 1 |

Other use case we have are emails that can be sent (by mail) and published to a webpage, were we pin the mail publication state, but the live publication can be changed later (for example because of typos)

mail
| id | text | publication_state | publication_id |
| 1 | xxx | draft | 0 |
| 2 | xxx | mail | 1 |
| 3 | xxx | live | 1 |

Marc-Roig:
>Agreed but I also agree with boegie here around implementing ES lifecycles too so each user can pick the right spot for their code to achieve what they want.

Yes! The current proposal is to create lifecycles on the Document Service layer, which also will be responsible to handle components.
In my opinion, having lifecycles on that layer will make the DB lifecycles much less necessary for people (but we will keep them neverhteless)

bglidwell:
Please keep and even further develop (by adding db support for things like cascade delete) the concept of "owned" records. ie. OneToMany OneToOne  

These are critical structural distinctions that can easily be overlooked when pursuing versioning/i18n/etc 

I am worried too much will be abstracted to the JS/TS side in strapi and the databases (and the data they store) will be useless outside the strapi application. Keep in mind several strapi devs are working on applications that store more than just marketing copy

PR #55, opened 2023-09-08
[V5] Plugin CLI

joshuaellis:
Introduce a dedicated plugin CLI to assist developers in creating plugins for strapi the right way easily

You can read it [here](https://github.com/strapi/rfcs/blob/v5/plugin-cli/rfcs/0055-plugin-cli.md)

### DIFF
--- a/rfcs/0055-plugin-cli.md
+++ b/rfcs/0055-plugin-cli.md
@@ -1,0 +1,69 @@
+- Start Date: 2023-09-08
+- RFC PR: https://github.com/strapi/rfcs/pull/55
+
+# Summary
+
+In v5, we want to stop merging plugins into the Strapi application before they are bundled. We instead want to bundle each plugin individually before it is imported, just like for any other dependency. This change will bring several benefits:
+
+- **Avoiding breaking changes:** When managing internal Strapi dependencies, we won’t have to worry about the side effects it could have on plugins that rely on those same dependencies. We’ll therefore be able to move quicker and keep up with the ecosystem.
+  - **Example: design system updates**
+    On the design system, we often want to do breaking changes, in order to improve the APIs. Doing this would require shipping a new major version. We own the Strapi Admin codebase, so we would be able to change the version of its dependency on `@strapi/design-system`, and do the manual changes required by the new APIs.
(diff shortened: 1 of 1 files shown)


ComfortablyCoding:
> Posting here from discord per request from joshua.

This looks great!

# Suggestions
One additional command that I would suggest is a db migration generation one. Enabling devs to generate knex migrations that will be picked up and executed at boot.

# Questions
For me the last major pains for plugins are:
1. Testing
    - When trying to test manually I need to install and have it within a strapi application which is not ideal, is it possible we would be able to do some sort of playground like nuxt modules can do? Is this what the link command is for?
    - Automated tests (both front-end and back-end) are not easily done. I assume that's why most plugins do not have any. 
    -  When testing the admin the errors are not descriptive or easy to find as build is awlays for prod.
3. Extensibility of core
    - A lot of internal but useful things (utils, components, hooks etc) are not exported and easily consumable by a plugin. It forces the devs to re-implement duplicate logic.

Any chance these are being addressed/looked into as well?

joshuaellis:
Thanks for the feedback @ComfortablyCoding.

# Testing

the `link` command should easy the ability for testing as you can use your local plugin as if it were installed in a strapi app, your set up could look like:

```shell
my-plugin/
├─ demo/
├─ src/
│  ├─ index.ts
├─ package.json
```

Where `demo` is a local strapi app – alternatively if you're building a plugin for your strapi app you could link it anywhere on your pc – we'll be using `yalc` to facilitate this. RE. automated testing, I agree, it's not easy to test strapi-fe at the moment externally unless you have a lot of deep knowledge. From the work we're doing in V5 a small ambition is to export some `test` utils either from the admin-package or elsewhere so you have a testing harness e.g. a store built etc. Finally regarding the build, we're aware the source-maps don't exist and are not ideal, this is something else we want to address 😄 

# Extensibility of Core

It'd be great to understand what you feel you're missing on this side so we can look into providing these exports. I think the naming conventions in the codebase at the moment are not very user friendly and i'd like to look at exporting more for users with clear names and guides.

ComfortablyCoding:
Nice, hopefully that resolves some issues.

In terms of the extensibility of the code it was more for the admin panel than the design system I guess. Technically unrelated here.

In terms of the design system their are common UI interfaces missing from the design system but present else where. For example the [components exported from the helper plugin](https://github.com/strapi/strapi/blob/main/packages/core/helper-plugin/src/index.js#L3).

Possibly already addressed by https://github.com/strapi/design-system/issues/1269.

PR #56, opened 2023-09-19
[V5]: FE APIs

joshuaellis:
Introduce typed imperative APIs to inject `DocumentActions`, `SideBarPanels` and `DocumentHeaderActions` as opposed to `InjectionZones` in V5.

You can read it [here](https://github.com/strapi/rfcs/blob/v5/fe-apis/rfcs/0056-FE-APIs.md)

### DIFF
--- a/rfcs/0056-FE-APIs.md
+++ b/rfcs/0056-FE-APIs.md
@@ -1,0 +1,234 @@
+- Start Date: 2023-09-19
+- RFC PR: https://github.com/strapi/rfcs/pull/56
+
+# Summary
+
+The following document proposes new public facing APIs aimed at plugin developers to be able to augment the `EditView` of the content-manager, the need came from the design proposals of Draft & Publish and would most likely be implemented only in V5 of the product under the Draft & Publish scope.
+
+This document tries to perform the minimum required work to retain backwards compatible functionality yet deliver a solid developer experience and pave the way for more productive and intuitive public facing FE orientated APIs.
+
+# Example
+
+```jsx
+// admin/src/index.ts
+import type { DocumentAction } from "@strapi/admin/admin";
+import { PublishMultipleLocales } from "./documentActions";
+
+export default {
+  bootstrap(app) {
+    app.addDocumentActions([PublishMultipleLocales]);
+
+    app.addEditViewSidePanels([
+      {
+        title: "Review Workflows",
+        content: <ReviewWorkflowsSidePanel />,
+      },
+    ]);
+
+    app.addDocumentHeaderAction([
+      {
+        label: "Select Locale",
+        options: [{ label: "english", value: "en" }],
+      },
+    ]);
+  },
+};
+
+// admin/src/documentActions.ts
+export const PublishMultipleLocales = {
+  label: "Publish multiple locales",
+  onClick: () => window.alert("you're publishing!"),
+};
+```
+
+# Motivation
+
+Current injectionZones let you inject components into areas of the content-manager, while this is a good it's a somewhat limiting API, the `addDocumentHeaderAction` & `addEditViewSidePanels` APIs while the former may seem very constricted provides a balance between user's being able to add actions quicker but not affect the design of the product, the latter is incredibly flexible but a much clearer typed system compared to the current implementation of `InjectionZones`. The `addDocumentActions` API is a new API that allows you to add actions to be rendered in the DocumentActions menu (this is new to V5).
+
+Overall, we want to move towards more developer friendly APIs that are quicker to use, more intuitive yet do not cause concerns for the usability of our product.
+
+# Detailed proposal
+
+> ⚠️ All types are rough and have not been validated, there may be reason to loosen types based on design specs.
+
+We’ll add the new APIs to the `app` class to keep with the current imperative model of introducing things to the Strapi admin panel. These APIs will be overloaded so you can apply the addition so it’s always rendered, or you can conditionally apply it based on the context provided.
+
+All the APIs dedicated to the EditView of the content manager will receive similar context payloads should user’s opt in to use the function overload, it can be illustrated like this:
+
+```tsx
+interface EditViewContext {
+  activeTab: "draft" | "published"; // room for more.
+  id: string;
+  meta: DocumentMeta;
+  document: Document;
+}
+
+// TBD – placeholder type
+interface DocumentMeta {
+  publishedAt: string;
+  createdAt: string;
+  createdBy: string;
+  // other dimensions e.g. locale
+  [key: string]: any;
+}
+
+// TBD – placeholder type
+interface Document {
+  // User attributes
+  [key: string]: any;
+}
+```
+
+We can then apply it to the APIs in a uniform fashion which creates consistency:
+
+```tsx
+interface ContentManagerEditViewAPIs {
+  funcName<TConfig extends object>(config: TConfig[]): void;
+  funcName<TConfig extends object>(configFn: ConfigFn<TConfig>): void;
+}
+
+type ConfigFn<TConfig extends object> = (
+  prevConfigs: TConfig[],
+  ctx: EditViewContext
+) => TConfig[];
+```
+
+Because the function receives the previous list of configs, we expect the complete list of configs to be returned, this means if a plugin required it could remove a configuration, or mutate a current configuration building upon one another e.g. the publish action would be able to be mutated by the `i18n` plugin correctly.
+
+## **Injecting into the Header**
+
+A `HeaderAction` is a constrained API enabling users to inject a specific set of items into the header toolbar section e.g. a `Select` or `Button`. It’s specifically restricted and we can make it looser as requirements surface and are validated.
+
+```tsx
+interface StrapiApp {
+  addDocumentHeaderAction(action: HeaderAction[]): void;
+  addDocumentHeaderAction(actionFn: HeaderActionFn): void;
+}
+
+type HeaderActionFn = (
+  previousPanel: HeaderAction[],
+  ctx: EditViewContext
+) => HeaderAction[];
+
+type HeaderAction = ButtonAction | SelectAction;
+
+interface ButtonAction {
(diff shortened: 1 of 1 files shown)


Boegie19:
Can you give us examples of how the injection zone replacement would be UI wise?

Boegie19:
I like the idea but why remove the InjectionZones and not build this feture ontop of InjectionZones. so it can be an InjectionZones or documentActions, SideBarPanels and DocumentHeaderActions

rfcs/0056-FE-APIs.md
```diff
@@ -0,0 +1,234 @@
...
+
+# Alternatives
+
+The alternative would be to use the current InjectionZones pattern more freely, but that has very little type safety and the pattern is essentially an uncontrolled closed-box where you “just pass props” and assume it goes to the right component leading to hard to test circumstances and unpredictability which imo makes for bad software. Product has also had concerns over the viability of the current InjectionZones pattern...
+
+# Unresolved questions
+
+- Do we want to allow DocumentActions to be written as if they’re react components but return a data structure? This allows actions to be even more powerful, but their implementation in Strapi becomes potentially more complex.
+- Should we live with injection zones in V4 and only deprecate them or completely remove them and do the additional work to define new APIs for the affected areas?
```
svrin, line 234: Please do not remove injection zones, but allow a graceful upgrade to the new api.

rfcs/0056-FE-APIs.md
```diff
@@ -0,0 +1,234 @@
...
+}
+
+type HeaderActionFn = (
+  previousPanel: HeaderAction[],
+  ctx: EditViewContext
+) => HeaderAction[];
+
+type HeaderAction = ButtonAction | SelectAction;
+
+interface ButtonAction {
```
Boegie19, line 115: we need more controls.
- control of the color of the button with. 
- permission support
- rbac conditional support
- visibility controls

rfcs/0056-FE-APIs.md
```diff
@@ -0,0 +1,234 @@
...
+  addDocumentAction(action: DocumentAction[]): void;
+  addDocumentAction(actionFunction: DocumentActionFunction): void;
+}
+
+type DocumentActionFunction = (
+  previousActions: DocumentAction[],
+  ctx: EditViewContext
+) => DocumentAction[];
+
+interface DocumentAction {
```
Boegie19, line 171: same as the Button action

rfcs/0056-FE-APIs.md
```diff
@@ -0,0 +1,234 @@
...
+  label: string;
+  onClick: (
+    event: React.MouseEvent<HTMLButtonElement>,
+    ctx: EditViewContext
+  ) => void;
+  icon?: React.ElementType;
+  disabled?: boolean;
+  dialog?: DialogOptions | NotifiationOptions | ModalOptions;
```
Boegie19, line 179: NotifiationOptions makes no send how do I know what notification to send if it happens on interaction with the button I don't even know if what I tried failed or succeeded NotifiationOptions should be a function that returns a notification

rfcs/0056-FE-APIs.md
```diff
@@ -0,0 +1,234 @@
...
+};
+```
+
+# Motivation
+
+Current injectionZones let you inject components into areas of the content-manager, while this is a good it's a somewhat limiting API, the `addDocumentHeaderAction` & `addEditViewSidePanels` APIs while the former may seem very constricted provides a balance between user's being able to add actions quicker but not affect the design of the product, the latter is incredibly flexible but a much clearer typed system compared to the current implementation of `InjectionZones`. The `addDocumentActions` API is a new API that allows you to add actions to be rendered in the DocumentActions menu (this is new to V5).
+
+Overall, we want to move towards more developer friendly APIs that are quicker to use, more intuitive yet do not cause concerns for the usability of our product.
+
+# Detailed proposal
```
Boegie19, line 50: I am missing a information table in here so we can add stuff to this. https://gyazo.com/d40ae81220023684f1bc479fbd630b62

rfcs/0056-FE-APIs.md
```diff
@@ -0,0 +1,234 @@
...
+  funcName<TConfig extends object>(configFn: ConfigFn<TConfig>): void;
+}
+
+type ConfigFn<TConfig extends object> = (
+  prevConfigs: TConfig[],
+  ctx: EditViewContext
+) => TConfig[];
+```
+
+Because the function receives the previous list of configs, we expect the complete list of configs to be returned, this means if a plugin required it could remove a configuration, or mutate a current configuration building upon one another e.g. the publish action would be able to be mutated by the `i18n` plugin correctly.
```
Boegie19, line 96: I see 1 issue with this implementation if there are no controls for the order the configs are read in how can I chance something that i18n changed if it is loaded after my function is done.

Boegie19:
We also need an easy way for plugins to add the special zones you made to the plugin like you can do with injection zones.

rfcs/0056-FE-APIs.md
joshuaellis: Do you? I think if you want all those things you should be using the `addDocumentHeaderAction(actionFn: HeaderActionFn): void;` so you could do something like:

```ts
addDocumentHeaderAction(() => {
  const { canRead } = useRBAC()

  if(!canRead){
    return null
  }

  return {}
})
```

I think the disabled property is probably helpful for sure, the color of the button, im not sure – that's a product decision it could be viable.

rfcs/0056-FE-APIs.md
joshuaellis: there's definitely use-cases e.g. your document action is to sync with an external data source, you just want a notificaiton that says "synced"

rfcs/0056-FE-APIs.md
Boegie19: @joshuaellis Lets say my use cease would be syncing document with something external then I would like to have a green button if fully synced orange when syncing and red when it is not synced it would allow you to communicate with the end user without them having to press the button first.

RBAC makes sence still the pemissions field like for side bar should exist

rfcs/0056-FE-APIs.md
Boegie19: @joshuaellis it would always be synced or error. if it says synced while it has an error that would not be a good user experience.

or can dialog also be a function?

t-fritsch:
Thanks for this RFC, 
V4 was a **_huge_** step back regarding frontend customization, compared to the freedom offered in V3. 
If I understand it well, this RFC propose a new way to *add* components in *some* pre-defined places (especially the header and edit view). 
This is great but V5 should also provide solutions to not only "add" stuff, but also affect **existing** parts of the UI :
- in the list view
- in the menu
- in the dashboard
- in popups
- etc.

There are numerous issues/posts about this : 
- https://github.com/strapi/strapi/issues/16873#issuecomment-1572227364
- https://github.com/strapi/strapi/issues/16873#issuecomment-1570507833
- https://github.com/strapi/strapi/issues/13523
- https://github.com/strapi/strapi/issues/13430#issuecomment-1144253709
- https://github.com/strapi/strapi/issues/13430#issuecomment-1146383361
- https://feedback.strapi.io/feature-requests/p/customize-the-admin-dashboard-welcome-page-v4
- https://feedback.strapi.io/feature-requests/p/html-customization-for-admin-panel-pages
- https://forum.strapi.io/t/extending-a-plugins-admin-interface/18907
- https://discord.com/channels/811989166782021633/1174689451281223710
- ...
I'd like to add that a lot of pain points that are raised could easily be solved by the issue-reporters themselves if strapi allowed customizing UI (meaning you'd get less issues and more community-made workarounds)

What is so frustrating with v4 is that the backend part can be customized quite fully : with routes, controllers, middlewares, services, we can do (almost) ANYTHING. This is perfect 🙌 
And when you switch to the frontend part you only get a tiiiiiiny space to play with (and always adding stuff, not editing or removing). It's like switching from paradise to jail 😰😅 

Please don't forget that if the developers that work with strapi are good enough to produce good backend code, we are also capable of doing some solid React Frontend development. 
I understand that it won't be possible in the very first V5 releases to match FE customization possibilities with the level of customization given BE-side, but please try to give us some freedom back 🙏

joshuaellis:
> Please don't forget that if the developers that work with strapi are good enough to produce good backend code, we are also capable of doing some solid React Frontend development.

No one's forgetting that, no one is saying that developers can't write frontend code. The issue primarily is it creates an unstable product where even the smallest changes we make can break other people's admin panels – you see this *consistently* with issues on the BE already. So what we're doing is weighing up customisation with stability because not every Strapi user modifies their product to such a high degree.

Hopefully, should these APIs be well received we can move to exposing more in the future because what we're doing is giving the ability to customise without compromising usability which is products core concern that I sympathise with.

Thank you for all your feedback though, it's well appreciated & we'll keep it all in mind!

t-fritsch:
Thanks for your quick reply @joshuaellis ! I realize you maybe misunderstood my intentions : I wasn't trying to accuse anyone of anything. Sorry if I wasn't clear. 🙏

I'm just emphasizing that there is a HUGE gap between the liberty you give BE side and FE side, and this cannot be solved by locking FE developers on so small areas like injections zones (If I understood the RFC the new system is quite the same). 
I think we can all imagine the frustration it would raise if you did the same thing BE side ie. allow developers to add stuff only in small places of the API or BE code, but not to override/remove core things. This is exactly the same on frontend.

I also disagree when you say that "_not every Strapi user modifies their product to such a high degree._". 
I think that modification of the dashboard for example is needed by almost every user, the fact that the majority of nowadays strapi users accomodate to that doesn't mean they don't need it. Same can be said about the menus, the default rendering of relations in list view, or the way dates are formatted, etc. 

I understand that this is important to not break things on every update and I've not digged deeper enough in strapi core to know what the solution would be, I'll leave it to the strapi team, but I hope you'll find a way to give back to developers better control of the UI, and not only on empty places but on existing part of the UI 🫶

PR #58, opened 2023-11-08
feat: add image focal point picker capabilities

federicobadini:
**RFC**: https://github.com/iFixit/strapi-rfcs/blob/main/rfcs/xxxx-focal-point.md
**Diff**: https://github.com/strapi/strapi/compare/main...iFixit:strapi:main

| Scenario | Image |
| --- | --- |
| Focal point mode trigger |  Screenshot 2023-11-08 at 17 48 45 |
| Focal point selection | Screenshot 2023-11-08 at 17 49 03 |
| Focal point selected | Screenshot 2023-11-08 at 17 49 18 |

### DIFF
--- a/rfcs/xxxx-focal-point.md
+++ b/rfcs/xxxx-focal-point.md
@@ -1,0 +1,38 @@
+- Start Date: 2023-11-08
+- RFC PR:
+
+# Summary
+
+We propose introducing a feature in Strapi that enables users to set a focal point on images directly from the media gallery.
+
+# Motivation
+
+As images are often utilized in resizable contexts, the absence of focal point information hinders the developer's ability to control cropping while preserving the most meaningful portion of the image. This has lead to content editors having to create multiple images, or designing a single piece of artwork with a centered subject that gets cropped on the sides.
+
+Several members of the community expressed interest in this feature too: see [here](https://feedback.strapi.io/feature-requests/p/image-focal-point)
+
+# Detailed design
+
+We suggest the addition of a focal point picker to the `EditAssetDialog` component. This implementation could draw inspiration from Strapi's existing image cropping interface.
+
+The `ActionRow` of the edit media dialog will have an added button that activates the set-focal-point mode. Upon activation, a viewfinder will appear for selecting the focal point on the image preview.
+
+Upon confirmation of the focal point position, these coordinates will be stored within the file's data structure as a JSON object featuring x and y percentage values.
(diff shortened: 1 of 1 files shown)


MihaiSandor:
Hi everyone, @federicobadini thanks for putting this RFC together.

CC @Bassel17 @derrickmehaffy @alexandrebodin @pierreburgy 

+1 from me. Our company needs this feature in production and it is becoming a release level blocker for us. We have content heavy properties where the same image is reused across many breakpoints and placements like hero banners, cards, and social previews. Without a focal point or safe area per asset, the generated crops often cut off the subject and editors are forced to re upload multiple pre cropped variants, which does not scale and creates a messy asset library.

If possible, we would really like to see this included in the next release cycle. Even a first version that supports setting a focal point on the asset and having image transformations respect it would immediately unlock a lot for teams like ours.

douwepausma:
This would be huge, we desperately need this feature. Our content-type schemas will otherwise be cluttered with fields to select focal point.

Schero94:
@douwepausma @MihaiSandor only review required ;)

pierreburgy:
Sorry everyone for the late reply. Thank you @Schero94 for this amazing contribution!

federicobadini:
Great to see this feature moving closer to Strapi core — and thanks to @Schero94 for picking this up and delivering an updated PR based on the original proposal.

One small note: we’re not sure the current contribution guidelines still match how the process works in practice. We went through the RFC route two years ago, but it didn’t lead to any discussion or follow-up. Since the guidelines still present RFCs as the recommended path for changes that impact core or all users, a bit of clarification could help future contributors avoid that path if that's no more the way to go.
There is an open issue to clarify the point: https://github.com/strapi/strapi/issues/25061

Looking forward to seeing this land! 🎯

Schero94:
@federicobadini tbh i didn't pick it, i asked @pierreburgy and i asked him, give me something to do this was one idea of him, so i said well let's do it! 

😁

pierreburgy:
@federicobadini I must admit we've haven't always led by example when it comes to community contributions (including pull requests reviews). The size of our team is limited but, as mentioned [here](https://www.linkedin.com/posts/pierre-burgy-strapi-88671673_bye-2025-hello-2026-i-have-something-to-activity-7416853178987634689-uZ7W), we want to do more with the community this year. 

@Schero94 you deserve all credits :)

